PR #5, opened 2017-07-26
Add favicons

hursey013:
Centralizing all favicon assets


src/favicons/browserconfig.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<browserconfig>
+  <msapplication>
+    <tile>
+      <square150x150logo src="/mstile-150x150.png"/>
```
zachmargolis: Do we need to make this a `/favicon/` url?

hursey013: I think that would only be necessary if we wanted this to work on the style-guide site, but since this will ultimately be extracted to the root of `identity-site` I think this makes sense.

PR #6, opened 2017-07-27
Remove documents folder

hursey013:
This is now in `.gitignore` so removing from repo.


zachmargolis:
Wait did the PDF get moved into the styleguide?

hursey013:
Yeah @gemfarmer had moved it in previously, likely with the intent to centralize all assets. I'm not sure that this PDF would probably ever be shared with the app, but may there is a use case for having PDFs like these cataloged in the style guide? What do you think?

zachmargolis:
I think this particular PDF, I'd leave in the marketing site only (since if we did reference it in the IDP, we'd just link to the marketing site anyways)

hursey013:
Yeah that makes sense, I can sort it out in another PR.

zachmargolis:
sounds good! thanks

PR #10, opened 2017-08-01
Language selector correction

hursey013:
My previous commit had some CSS that was overly greedy and impacted the mobile help dropdown menu... this fixes it 😬


zachmargolis:
LGTM

PR #13, opened 2017-08-09
Added special case for spanish menus to avoid wrapping

nickbristow:
**Why**:
To prevent text wrapping at specific screen dimensions


nickbristow:
https://github.com/18F/identity-private/issues/2262

nickbristow:
@hursey013 done

PR #14, opened 2017-08-11
Adjusted margin on ES nav

nickbristow:
**Why**:
Margins were still too tight and caused wrapping

### DIFF
--- a/src/css/components/_nav.scss
+++ b/src/css/components/_nav.scss
@@ -89,6 +89,9 @@ header nav {
 
 @media #{$breakpoint-es}{
   header nav.es-nav {
-    font-size: $sm-es; 
+    font-size: $sm-es;
+    ul li{
+      margin-right: 1.5rem;
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


src/css/components/_nav.scss
```diff
@@ -89,6 +89,9 @@ header nav {
 
 @media #{$breakpoint-es}{
   header nav.es-nav {
-    font-size: $sm-es; 
+    font-size: $sm-es;
+    ul li{
```
gemfarmer: nit pick: space between `li` and `{`. It might be worth adding hound on this repo so that someone doesn't have to comment on syntax :man_shrugging:

adunkman:
Thanks for the pull request, @nickbristow! 

There’s currently a project undergoing to bring the style guide into greater usage (including an update to use [USWDS 2.0](https://github.com/uswds/uswds), which includes some big code changes here — and as a result this PR is out of date, and I’ll close it to keep things tidy for this repo as well as for your open PR list. 💯  

Be sure to reach out if you have any questions!

PR #20, opened 2018-12-27
Add base Buttons page, styles.

adunkman:
This PR introduces button styles and the corresponding documentation page.

A few notes/questions for @rtwell:

- I noticed the button styles in the Sketch document didn’t have active styles, so I improvised. I think the active styles are important to have, but feel free to adjust or provide new colors. 
- Additionally, the colors used for the buttons in the Sketch document don’t match the color palette provided in the same document. I preferred the color palette colors over the colors used in the button mockups, but let me know if that was incorrect (and if so, the desired colors). 

More generally:

- I feel like I’m missing something on how I’m supposed to specify custom colors to USWDS. Maybe @maya has some insight here, or can point me to the relevant documentation?
- I moved the section on Radio Buttons to the Form Fields page. That felt like where I’d go looking for it, and mimics the decision made by [the USWDS documentation](https://v2.designsystem.digital.gov/components/form-controls/#radio-buttons).
- I plan to introduce accessibility testing shortly, specifically to catch contrast issues — I’m worried about those disabled styles.
- The content (perhaps obviously) is lacking, and there feels like there’s a distinction between the first four button types and the last two, but I don’t really know without knowing more about their usage. I think maybe @donjo knows more perhaps?

[Screenshot of buttons documentation page](https://federalist-proxy.app.cloud.gov/preview/18f/identity-style-guide/buttons/components/buttons/)


donjo:
> I plan to introduce accessibility testing shortly, specifically to catch contrast issues — I’m worried about those disabled styles. 

Ryan may have more clarity here but the guidance we were going on for the disabled state buttons states that inactive user interface elements are exempt from contrast ratio requirements so we went with this style to make it clear to users that these buttons are not clickable [Understanding SC 1.4.3
 on W3.org](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html)

> The content (perhaps obviously) is lacking, and there feels like there’s a distinction between the first four button types and the last two, but I don’t really know without knowing more about their usage. I think maybe @donjo knows more perhaps?

I'll need to come up with some guidance here but the preview answer is these buttons are for making changes to content on the page. Mostly used on the account page right now for editing your phone number and stuff like that.

> I moved the section on Radio Buttons to the Form Fields page. That felt like where I’d go looking for it, and mimics the decision made by the USWDS documentation.

👍 

I'll look into the custom colors stuff a bit later but wanted to get the answers I have out there sooner than later.

maya:
>I feel like I’m missing something on how I’m supposed to specify custom colors to USWDS. Maybe @maya has some insight here, or can point me to the relevant documentation?

It supports changing theme colors from system colors, and also turning on different system colors.

Relevant links:

- https://v2.designsystem.digital.gov/utilities/color/
- https://v2.designsystem.digital.gov/design-tokens/color/theme-tokens/
- https://v2.designsystem.digital.gov/design-tokens/color/system-tokens/
- https://github.com/uswds/uswds/blob/release-2.0/src/stylesheets/theme/_uswds-theme-color.scss 

It doesn't inherently support colors outside of the system palette as is designed in Login.gov, so you'd probably have to override the theme variants and color maps rather than setting a color family in the theme file.

rtwell:
Hey @adunkman!

Thanks for all of your hard work on this—it's so great to see this coming to life, and so quickly!

To answer a couple questions you had:

1) I _think_ the only color that was a mismatch in the sketch file was the middle-green tone. The one I left in the sketch file is too light and won't be 508 compliant with white text over it. I've uploaded a new sketch file `login-gov_styleguide_v1.9.3.sketch` with the proper green. The new green is `#18852E`

2) The `active` colors! here they are:
     • Primary `#112E51` _(already in the sketch file as **Navy**)_
     • Secondary `#EBF3FA` _(already in the sketch file as **Light Blue**)_
     • Success `#116523` _(already in the sketch file as **Dark Green**)_
     • Warning `#AB142F` _(already in the sketch file as **Dark Red** )_
The new sketch file also include these buttons.

Thanks again for your help and please let me know if you have any questions!

adunkman:
Thanks for the feedback, all! Things are getting closer — amended the commit with the updated colors, and updated the screenshot above. You can always see the latest on the [Federalist preview](https://federalist-proxy.app.cloud.gov/preview/18f/identity-style-guide/buttons/components/buttons/).

> … the guidance we were going on for the disabled state buttons states that inactive user interface elements are exempt from contrast ratio requirements…

Got it! That makes sense. #21 will put our mind at ease for future accessibility-related questions — once that’s merged, we’ll fail builds on accessibility issues.

> It doesn't inherently support colors outside of the system palette as is designed in Login.gov, so you'd probably have to override the theme variants and color maps rather than setting a color family in the theme file.

Got it! I’ve taken that approach in this PR — glad I wasn’t missing something! 👍 

@rtwell Looks like there might be some color discrepancy still — there are two colors used in the Sketch file (v1.9.3) which aren’t in the color palette: 

- The success button’s `active` state, which appears to be `#094316` (`#116523` **Dark Green** is the button’s hover color).
- The danger button’s `active` state, which appears to be `#8B0A21` (again, `#AB142F` **Dark Red** is the button’s hover color). 

Should we add those (or similar) colors to the palette, or go without active states for those buttons? (The latest commit here goes without active states, you can see what it feels like at the above Federalist preview URL).

Thanks, all! Once the colors are settled, I’ll look for a 👍 from one of y’all to merge — and we can do a follow-up PR with the guidance on how the different button types are used, if that’s okay with y’all.

rtwell:
awesome, thank you!

adunkman:
As per discussion in Slack with @rtwell, the above colors should be added to the palette (`#094316` and `#8B0A21`), but we haven’t yet settled on names for those colors (we may rename them all as well). 

Just to get this pull request moving, I’ll call them Darkest Green and Darkest Red for now, and we can circle back to rename the SASS variables when the time comes. I’ll add a commit here, and merge on green build. 🚀

PR #28, opened 2019-01-10
Add Integration tests, cover Color and Accordion pages.

adunkman:
This introduces integration tests (running in a headless version of Chrome via [`puppeteer`](https://github.com/GoogleChrome/puppeteer)) to assert that:

1. the colors written on the Color page in hex match the background color of the swatch they represent (asserting the documentation matches the related Sass files), and
2. the accordions on the Accordion page behave as expected.

These should help us maintain confidence when performing upgrades to `uswds` and other dependencies. 🤘

I’m going to reach out to `#g-frontend` for a review as well, since I’m not sure we have precedent for complete integration tests in JavaScript (or if so, we should add ‘em to [the guide](https://github.com/18F/development-guide/tree/master/nodejs)).


docs/_includes/helpers/code-example.html
```diff
@@ -1,5 +1,5 @@
 {{ include.code }}
 
 ```html
-{{ include.code }}
+{{ include.code | replace_regex: '\s?test-[^\s"]+', '' }}
```
adunkman: This removes `.test-*` classes from code samples, as they’re used for integration tests but aren’t relevant for consumers of the documentation.

(Using `.test-*` classes makes selecting targets in tests much more durable).

test/color.test.js
```diff
@@ -0,0 +1,24 @@
...
+  const [, red, green, blue] = /rgb\((\d+), (\d+), (\d+)\)/.exec(rgb);
+  const parts = [red, green, blue].map(d => `0${Number(d).toString(16)}`.slice(-2));
+  return `#${parts.join('')}`;
+};
+
+test('hex codes match background color on the color documentation page', async () => {
+  await page.goto(`${host}/color/`);
+  const colors = await page.$$eval('.test-color-swatch', els => [...els].map((el) => {
+    const text = el.innerText.trim();
+    const background = window.getComputedStyle(el).backgroundColor;
```
adunkman: `getComputedStyle()` returns colors in the form of `rgb(255, 255, 255)`. I don’t believe there’s a way to get it to return hex colors, hence the conversion function above.

test/.eslintrc.yml
```diff
@@ -0,0 +1,13 @@
+env:
+  jest: true
+globals:
+  afterAll: true
```
davemcorwin: doesn't adding `jest: true` under the `env` key include all the testing globals such as `afterEach`?

test/components/accordions.test.js
```diff
@@ -0,0 +1,49 @@
+const host = `http://localhost:${process.env.JEST_PORT}`;
+
+beforeEach(async () => page.goto(`${host}/components/accordions/`));
+
+test('accordions are open only when aria-expanded="true"', async () => {
+  await expect(page).not.toMatchElement('.test-visible[hidden]');
```
davemcorwin: for context, should it be asserted that `aria-expanded` is not `true`? Also, why not just use the id selectors as in the subsequent tests? If you're checking `aria-expanded` which s scoped to each accordion container, it may be clearer.

docs/_includes/helpers/code-example.html
adunkman: I think it’s about the same in terms of risk of collision — the `data-test="name"` attribute can also be styled (but why would you?). I think the same rationale could be used for applying styles to `.test-name` classes.

That being said, I’m not married to any particular decision — the goal here was to be able to clearly select elements for testing purposes with selectors that aren’t brittle (for example, using `.test-accordion` instead of `.usa-prose .usa-accordion` (there’s also an accordion in the menu bar).

test/.eslintrc.yml
adunkman: I’ll do some more investigation here. I believe they were required — but I’ll see if I can remove it. If they’re required, I’ll figure out what `env: jest: true` does… perhaps it’s meaningless.

adunkman: Good call! They are not required — removed in 03e2aa0.

docs/_includes/helpers/code-example.html
adunkman: Switched to `data-test="name"` in 063a2d1 — I don’t have a strong preference, so I’ll go forward with your suggestion!

test/components/accordions.test.js
adunkman: > for context, should it be asserted that `aria-expanded` is not `true`?

Great idea! Added in 063a2d1, and did a bit of test refactoring to reduce duplicating that everywhere. 

> Also, why not just use the id selectors as in the subsequent tests?

I was doing so to reduce the number of assertions, but that’s silly (asserting `.test-hidden[hidden]` is actually acting on 2 IDs). I dropped the test classes and asserted the IDs individually in 063a2d1.

adunkman:
@davemcorwin I believe I’ve addressed the above comments — could you swing by for a quick review again? Thanks for your help!

PR #30, opened 2019-01-17
Add basic code highlighting.

adunkman:
As mentioned in #27, this implements basic code highlighting styles, based off the login.gov color space. These are pygments compatible class names, which are the standard across every standard syntax highlighter.

Additionally, this PR applies the theme monospace font (Roboto Mono Web) to all `code` and `pre` elements for convenience and compatibility with default markdown syntax highlighting.

Here’s what it looks like on the Accordions page:

[Screenshot of Accordions documentation page](https://federalist-proxy.app.cloud.gov/preview/18f/identity-style-guide/code-styles/components/accordions/)

I didn’t spend an incredible amount of time on this, so open to changes and iterations as we discover what doesn’t work.

### DIFF
--- a/src/scss/components/_code.scss
+++ b/src/scss/components/_code.scss
@@ -1,0 +1,73 @@
+// stylelint-disable declaration-block-single-line-max-declarations
+
+pre,
+code {
+  @include u-font('mono', $theme-body-font-size);
+}
+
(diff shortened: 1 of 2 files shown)


adunkman:
Actually, I think this should be covered in documentation as well, since at least developers.login.gov will be using these code styles. Created #31 to add documentation later.

PR #32, opened 2019-01-22
Implement form component styles.

adunkman:
This PR adds styles and documentation for form fields (text inputs, dates, selects, radio buttons, and checkboxes).

This causes a build failure, due to insufficient contrast on the disabled input labels. We’ll need to discuss what to do to resolve that — not sure if [we consider that a valid failure of accessibility](https://github.com/18F/identity-style-guide/pull/20#issuecomment-450381638).

[Screenshot of form fields documentation page](https://federalist-proxy.app.cloud.gov/preview/18f/identity-style-guide/form-fields/components/form-fields/)


docs/_components/form-fields.md
```diff
@@ -1,22 +1,571 @@
...
+    <div class="usa-form-group usa-form-group-day">
+      <label for="b857" class="usa-label">Day</label>
+      <input id="b857" class="usa-input usa-input-inline usa-input-error" aria-describedby="f9e5" type="number" min="1" max="31">
+    </div>
+    <div class="usa-form-group usa-form-group-year">
+      <label for="a5dc" class="usa-label">Year</label>
+      <input id="a5dc" class="usa-input usa-input-inline usa-input-error" aria-describedby="f9e5" type="number" min="1900" max="2000">
+    </div>
+  </div>
+  <span class="usa-input-error-message" role="alert">Error message text.</span>
```
donjo: I was thinking maybe we should include the error message icon that form fields use inside the field at the start of these error messages when they're paired with checkboxes/radios so text color isn't the only signal that something is wrong.

docs/_components/form-fields.md
```diff
@@ -1,22 +1,571 @@
...
+  <span class="usa-form-hint" id="f9e5">Example: 04 28 1986</span>
```
donjo: Also thinking that the error should appear on a line by itself so the hint necessary to validate the field is still aligned with the fields.

I'll say the error should come after the hint but open to suggestions as I don't really have a good example to go off of.

adunkman: I’ll go with after, I also think it seems to make more sense logically (the error state doesn’t break apart the existing control, which I imagine contains the hint). 

Here’s both options visually, just for reference. 

| Before hint | After hint |
|------------|----------|
| Screenshot of errored date field with error message showing before hint text | Screenshot of errored date field with error message showing after hint text |

docs/_components/form-fields.md
adunkman: Here’s what that would look like across the styles — the differences in padding and border create some interesting alignment problems. 

Small sized radios with error message

Normal sized radios with error message

Large sized radios with error message

Thoughts? It’s definitely more usable since it’s no longer just relying on color to indicate state, but the alignment gives me pause (and seems like it’d rarely be used correctly if we added three different classes for each of these situations). 

I see why the US Web Design System went with [a simple left border](https://v2.designsystem.digital.gov/components/form-controls/#text-input) to indicate this… seems versatile (as most forms align along a simple vertical axis).

docs/_components/form-fields.md
donjo: 👍  on after

docs/_components/form-fields.md
donjo: I feel like in the long run we may want to re-think this and come up with something more sustainable (like the left border) but in the mean time I made a small adjustment of `padding-left: 1.5rem` on `.usa-input-error-message.usa-input-error-message-with-icon {` and I think it looks a bit better since the icon and text isn't really lining up with the small inputs and it just looks like a sentence with an icon in front. Let's go with this for now to keep things simple.

screen shot 2019-01-30 at 10 54 15 am
screen shot 2019-01-30 at 10 54 01 am
screen shot 2019-01-30 at 10 53 47 am

donjo:
> This causes a build failure, due to insufficient contrast on the disabled input labels. We’ll need to discuss what to do to resolve that — not sure if we consider that a valid failure of accessibility.

After discussing this with some other visual designers it sounds like this may be "compliant" but not necessarily good. Is there an easy way to skip this test for disabled inputs as we work on some fresh button designs? I'll create a new issue for that.

src/scss/components/_inputs.scss
```diff
@@ -167,6 +167,12 @@ $border-width: 1px;
...
 
 .usa-input-error-message {
-  display: inline-block;
   color: color('error');
+
+  &.usa-input-error-message-with-icon {
+    @include add-background-svg('alerts/error');
+    background-position: 0 center;
+    background-size: units($icon-size);
+    padding-left: units($icon-size) + units(1);
```
donjo: `padding-left: units($icon-size) + units(1);`  >> `padding-left: units($icon-size);` is what I was referring to in the other comment

adunkman:
I believe this should be ready for review again!

donjo:
LGTM 👍

PR #46, opened 2019-01-24
Adds docs page for automatic syntax highlighting

el-mapache:
Here is a first pass at providing documentation for our syntax highlighting. I went with general information since the actual implementation depends on what language / framework are being leveraged to generate the highlighted syntax.

Hopefully the tone and writing style are consistent with what we currently have in the style guide!

Feedback + comments welcome


docs/_components/code-formatting.md
```diff
@@ -0,0 +1,86 @@
+---
+title: Code Formatting
```
adunkman: According to [the Content Guide](https://content-guide.18f.gov/capitalization/#headings), headlines, page titles, subheads and similar content should follow sentence case. 

(I think sentence-case heading look a little weird personally, but those are just my tastes! I think it’s important to be consistent.)

docs/_components/code-formatting.md
```diff
@@ -0,0 +1,86 @@
...
+lead: >
+  Facilitates syntax highlighting that maintains visually consistentcy with the login.gov color palette
```
adunkman: `consistentcy => consistency`

docs/_components/code-formatting.md
```diff
@@ -0,0 +1,86 @@
...
+---
+
+Intended for use by other developers when producing documentation for sites that leverage the login.gov styleguide,
```
adunkman: `styleguide => style guide`

docs/_components/code-formatting.md
```diff
@@ -0,0 +1,86 @@
...
+.s1 /* literal.string.single */
+.ss /* literal.string.symbol */
+.bp /* name.builtin.pseudo */
+.fm /* name.function.magic */
+.vc /* name.variable.class */
+.vg /* name.variable.global */
+.vi /* name.variable.instance */
+.vm /* name.variable.magic */
+.il /* literal.number.integer.long */
+```
```
adunkman: I’m concerned that this list of items will become out of date, should we make updates in the future to the scss file. We could write a test to assert these two lists are the same, but I think we might be able to drop the list instead — since these classes are written by highlighters and are a standard.

Thoughts?

docs/_components/code-formatting.md
```diff
@@ -0,0 +1,86 @@
...
+
+These classes should provide functional highlighting for any of the languages Pygments supports.
+For a complete list of tokens, and for more general information on Pygments, please refer to their
+documentation, [found here](http://pygments.org/).
```
adunkman: As per [the Content Guide](https://content-guide.18f.gov/urls-and-filenames/#link-text): 

> … users of screen readers will often skip from one link to another, skipping the text in between, as a way of skimming for the content they need. … Avoid ambiguous link text like click here, here, or learn more whenever possible.

Perhaps: 

> … for more general information on Pygments, please refer to the [Pygments documentation](http://pygments.org/).

docs/_components/code-formatting.md
el-mapache: Yes, a test for this seems like overkill. I'll just add a link to the CSS if people want to inspect which classes are supported with the login.gov color palette.

el-mapache:
@adunkman thanks! I've updated the PR with your requested changes.

PR #51, opened 2019-02-04
Fix issue with current navigation item highlighting; move menu.

### LINKED ISSUE #36: Write or delete the Components Overview page.
The [Components Overview page](https://federalist-proxy.app.cloud.gov/site/18f/identity-style-guide/components/) is currently blank. 

Is there anything that should go there, or should we delete it? 🔮

adunkman:
This PR contains a few navigation-related changes:

1. Fixes an issue where the current primary navigation item wasn’t selected when viewing collection pages. For example, when viewing the “Buttons” page, “Components” was not highlighted in the primary navigation as it should have been. 

    The navigation helpers were previously copied into this repo from [18F/uswds-data](https://github.com/18F/uswds-data), and I was having trouble understanding and debugging issues — so this PR makes changes to those helpers to make them more straightforward.

2. Deletes the blank page at `/components/overview` (to close #36). There doesn’t seem to be a need for it. 

3. Changes the header format to the Basic style. This moves the primary navigation to the right side of the header, dropping the Download link (which didn’t actually download anything). Additionally, I dropped the link to the homepage, as it’s also linked at the logo. Here’s what that looks like: 

    | Before | After | 
    |----------|------|
    | Screenshot with old navigation | Screenshot with new navigation |


adunkman:
This isn’t particularly risky, and since @donjo is out this week, I’m going to merge away — but if you’re coming across this and have questions or feedback, I’d be happy to hear it!

PR #61, opened 2019-02-11
Extract usage instructions to page; add plugin.

adunkman:
Adds the Jekyll plugin used in https://github.com/18F/identity-dev-docs/pull/67 as a reference to the usage instructions, and moves the usage instructions to its own page.

I’ll follow-up this PR with extended usage instructions for the **Compiling source files in your build process** section, after I’ve done a basic integration that does so.

[Screenshot of usage instructions](https://federalist-proxy.app.cloud.gov/preview/18f/identity-style-guide/documentation-updates/usage/)

### DIFF
--- a/docs/_data/nav.yml
+++ b/docs/_data/nav.yml
@@ -1,4 +1,5 @@
 primary:
+  - href: /usage/
   - href: /brand/
   - href: /typography/
   - href: /color/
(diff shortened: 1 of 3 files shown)


adunkman:
As this is a documentation-only update, I’ll merge — but feel free to give feedback here, I’m happy to receive it!

PR #62, opened 2019-02-12
Simplify usage of scss within Rails.

adunkman:
This PR makes extending the source Sass files easier within any project, and especially easy within Rails projects.

- The source Sass files are copied over to `dist/assets/scss`, including the `uswds` dependency. The paths used within the style guide to include `uswds` are rewritten to reference the local folder, so no dependency issues will appear when this library is used in another build pipeline.
- Anytime `url([something])` is encountered, it is wrapped in `url(asset-path-if-defined([something]))`. The `asset-path-if-defined()` function passes through the path unchanged, unless the `asset-path()` function is also defined (as it is [within the Rails Asset Pipeline](https://guides.rubyonrails.org/asset_pipeline.html#css-and-sass)).
- The `$font-path` and `$img-path` variables are now exposed, which control where fonts and images are referenced from (before they are passed into `asset-path()`, if it is defined). 

The result is a straightforward import from Sass, and transparent support for the Rails Asset Pipeline, as demonstrated in the documentation update:

[image](https://federalist-proxy.app.cloud.gov/preview/18f/identity-style-guide/src-include-improvements/usage/)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-dist
-node_modules
-tmp
+/dist
+/node_modules
+/tmp
 .DS_Store
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ build: build-docs build-assets
 build-docs:
 	JEKYLL_ENV=production bundle exec jekyll build
 
-build-assets: build-sass-and-js build-fonts build-images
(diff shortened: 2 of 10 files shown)


adunkman:
Going to merge this without review as it’s a build process + documentation update, but I’m always open to feedback should you have it! :heart:

PR #66, opened 2019-05-10
2.0 Update

donjo:
Update to USWDS 2.0!

Things that still need to be fixed...

- [x] Buttons 
- [x] Form Controls
- [x] Navigation
- [x] Accordions
- [x] Alert paragraph guidance
- [x] Footer logo size and positioning
- [x] Sample sites on the layout page


lauraGgit:
Looks like there might be a minor error on the tooltip hover css:
image

donjo:
Screen Shot 2019-05-17 at 12 24 33 PM

noticing that accordion headers that are closed look the same as ones that are open. I'll look into that. too along with the tooltip

package.json
```diff
@@ -65,7 +65,7 @@
     "serve": "^10.1.1",
     "stylelint": "^9.9.0",
     "stylelint-scss": "^3.4.1",
-    "uswds": "^2.0.0-beta.4",
+    "uswds": "^2.0.1",
```
lauraGgit: 🎉

PR #67, opened 2019-05-13
Fixing navigation

donjo:
Screen Shot 2019-05-13 at 12 50 45 PM
Screen Shot 2019-05-13 at 12 50 22 PM

This fixes the navigation so it displays properly but also cleans up the design to make it a little tighter.

### DIFF
--- a/src/scss/components/_navbar.scss
+++ b/src/scss/components/_navbar.scss
@@ -1,24 +1,23 @@
-$header-height: 15;
+$header-height: 10;
 
 .usa-header--extended,
 .usa-header--basic {
   .usa-logo {
     @include at-media-max($theme-navigation-width) {
       font-size: unset;
-      line-height: unset;
       flex: unset;
     }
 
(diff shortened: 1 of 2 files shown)


src/scss/components/_navbar.scss
```diff
@@ -118,13 +117,26 @@ $header-height: 15;
...
-  @include u-font-size($theme-header-font-family, 4);
+.usa-logo__text {
+  @include at-media-max($theme-navigation-width) {
+    @include u-font-size($theme-header-font-family, 3);
+    padding-left: 17px;
+  }
+
+  @include at-media($theme-navigation-width) {
+    @include u-font-size($theme-header-font-family, 4);
+    padding-left: 27px;
```
donjo: Kinda magic number usage here but we're trying to position text against an image where the logo and text in the logo is all one image. We should eventually re-build this logo in a more sustainable way

PR #68, opened 2019-05-13
2.0 templates

donjo:
This mostly fixes the page layout templates (on /layout) though some of the weird stuff you still see are dependent on buttons and accordions

Screen Shot 2019-05-13 at 5 07 13 PM

### DIFF
--- a/docs/examples/extended-header-no-sidenav-medium-footer.html
+++ b/docs/examples/extended-header-no-sidenav-medium-footer.html
@@ -6,41 +6,41 @@ layout: base
 
 <div class="usa-overlay"></div>
 
-<header class="usa-header usa-header-extended" role="banner">
+<header class="usa-header usa-header--extended" role="banner">
   <div class="usa-navbar">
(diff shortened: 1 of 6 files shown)


docs/examples/extended-header-no-sidenav-medium-footer.html
```diff
@@ -6,59 +6,59 @@
 
 <div class="usa-overlay"></div>
 
-<header class="usa-header usa-header-extended" role="banner">
+<header class="usa-header usa-header--extended" role="banner">
```
lauraGgit: in the docs it looks like its still `usa-header-extended`

lauraGgit: https://designsystem.digital.gov/components/headers/#extended

donjo: It looks like that's the original docs site (for 1.x). On the v2 docs it's still listed as `--extended`. Just to be sure I double checked and a single `-` breaks the nav

PR #70, opened 2019-05-14
2.0 forms

donjo:
Form fixes 

Some things on this page aren't fixed yet since they fall under the accordion CSS which we're addressing in a separate PR.

Screen Shot 2019-05-14 at 12 25 19 PM


donjo:
@lauraGgit I think that's just an effect of the accordions being broken and the content is being shown while normally it would be hidden by default. I'll circle back to this after the accordion PR is merged. 
Screen Shot 2019-05-14 at 4 56 48 PM

PR #74, opened 2019-05-16
2.0 pa11y fixes

achapm:
Here is what the disabled color that pa11y requires:
image

Here is the new teal that pa11y requires:
image


src/scss/uswds-theme/_color.scss
```diff
@@ -25,16 +25,16 @@ $site-palette: (
   'site-light-blue': #ebf3fa,
   'site-input-blue': #f2f9ff,
 
-  'site-teal': #00bfe7,
-  'site-light-teal': #ecfcff,
+  'site-teal': #00a2c4,
+  'site-light-teal': #00a2c4,
```
donjo: This makes light teal and teal the same color. I think we still want to use light teal as it stands in some places but use it in an accessible way. How/where was the test failing?

achapm: It was not causing a failure it seems. Reverted change to site-light-teal.

PR #79, opened 2019-05-29
update usage guidance

lauraGgit:
per 1162 - add a bit more guidance on the existing components.


docs/_components/buttons.md
```diff
@@ -63,7 +65,7 @@ Use these button styles unless doing x and y, in which case, use [other buttons]
 
 # Other Buttons
 
-These buttons should be used when x-ing and y-ing.
+These buttons if the actions are not among the top two actions in a view, use an other button to convey action.
```
donjo: "Use these buttons" or "These buttons should be used"

lauraGgit: ah good catch, updated.

PR #81, opened 2019-05-30
add badge code and guidance

lauraGgit:


src/scss/components/_verification-badge.scss
```diff
@@ -0,0 +1,26 @@
+.lg-verification-badge {
+
+  background-color: map-get($site-palette, 'site-white');
+  border: units(1px) solid map-get($site-palette, 'site-green');
+  border-radius: $theme-border-radius-xxl;
+  box-sizing: border-box;
+  display: inline-block;
+  padding: units(1) units(2);
```
donjo: USWDS has some globally available spacing units so I swapped the space variables to these. https://v2.designsystem.digital.gov/design-tokens/spacing-units/

docs/_components/badges.md
```diff
@@ -0,0 +1,33 @@
...
+Badges can help signal to the user that they had completed a workflow desired by Login.gov. Primarily, these are used on our user account page to encourage more secure practices.
+
+Badges are comprised with a div using the `lg-verification-badge` class and a small SVG image alongside a `span` of text.
+
+## An unphishable badge example
+
+The unphishable badge is used to indicate that an account is only using security keys as its MFA method.
+
+{% capture example %}
+<div class="lg-verification-badge">
```
donjo: I restructured this to be an img and span inside a div to make it easier to use the `vertical-alignment` controls.

src/scss/components/_verification-badge.scss
lauraGgit: ah nice!

docs/_components/badges.md
lauraGgit: ah that makes sense

PR #82, opened 2019-05-30
add spinner component

lauraGgit:


src/scss/components/_spinner.scss
```diff
@@ -0,0 +1,17 @@
+.lg-spinner {
+  margin-left: auto;
```
donjo: I think the spinner class itself shouldn't care about where it sits on the screen. While our current implementation has it there, we don't know if all future versions will put it in the same place.

docs/_components/spinner.md
```diff
@@ -0,0 +1,34 @@
...
+
+Spinners can be helpful to signal to a user that the current process may require a few seconds.
+
+**Before using a spinner** Consider if there is  a way to improve the performance of the sign to mitigate the use of a spinner.
+
+## Example of a spinner
+
+{% capture example %}
+<div class="lg-spinner">
+    <div>
```
donjo: What do you think about making these code samples 2 space indent instead of 4? It's what we go with for the Sass code and makes it a bit easier to read inside the code sample blocks

PR #83, opened 2019-05-31
Lg/layout typography help

lauraGgit:


docs/_components/footer.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Footer
+---
+
+Check out the slim and medium footer examples on the [layout page](/layout#page-layouts).
```
donjo: These links aren't quite working. May need to use some kind of site root variable in front of the `/`

lauraGgit: ok, added the baseurl - but we'll check on the federalist build

PR #101, opened 2019-08-20
Style guide layout polish

donjo:
Going through our docs and making sure we're consistent with our headers and other things to set a good foundation for new additions.


donjo:
### When there's a docs page with no sidenav it gets the big H1
Screen Shot 2019-08-23 at 11 20 08 AM

### When there's a sidenav the H1 is smaller 
Screen Shot 2019-08-23 at 11 19 54 AM

donjo:
Also upgraded Jekyll to 4.0 for faster builds

donjo:
### Code blocks are no longer 100% width
Screen Shot 2019-08-23 at 11 23 28 AM

Open to suggestions on this. The standard prose width seemed too narrow so I went with something a bit wider.

PR #110, opened 2019-09-06
LG-1480 Add guidance about using alerts

jmhooper:
This commit adds guidance about when to use alerts to the alert section of the design system

### DIFF
--- a/docs/_components/alerts.md
+++ b/docs/_components/alerts.md
@@ -1,6 +1,6 @@
 ---
 title: Alerts
-lead: |
+lead: >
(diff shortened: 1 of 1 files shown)


lauraGgit:
Might we also want to link back to the uswds alert guidelines? https://designsystem.digital.gov/components/alert/

donjo:
Thanks for submitting this @jmhooper - I made some small edits and moved it to the main body so it could be broken into paragraphs.

Thanks @lauraGgit for the suggestion, added that too.

PR #128, opened 2020-01-14
Lg 1810 Create styling for user actions on dashboard

karlaturcios:
Why? Because a UI control is needed so users can select action items from a drop down menu.


karlaturcios:
Screen Shot 2020-01-21 at 2 53 08 PM
Increased the contract for the active dropdown button. It's has a Contrast Ratio
7.67:1

karlaturcios:
Merged, @jmhooper please advice on cutting a new release, this will be my first one :)

Issue #102, opened 2019-08-20
Consider using CSS classes for icons in the design system


karlaturcios:
Would it be worthwhile to create CSS classes that would include the path to the image along with styles such as height and width declaration, and color? That way code could be easily maintained and updated in fewer places when using images.

karlaturcios:
LG-1600 is set for tracking

porta-antiporta:
tracking all in JIRA. closing.

PR #134, opened 2020-02-14
Patch _code.scss rendering 

nickttng:
**Why:** To patch a rendering issue identified in [identity-dev-docs' PR #102](https://github.com/18F/identity-dev-docs/pull/102)

**How:** Redefine both `.highlight::before` and `.highlight::after` width from `$block-border-width` to `0`

When reviewed and merged, we will publish a `patch` release to npm

## Before (left column) / after (right column) rendering for Chrome, Firefox and Safari

code-highlight-rendering

### DIFF
--- a/src/scss/components/_code.scss
+++ b/src/scss/components/_code.scss
@@ -44,7 +44,7 @@ pre {
     z-index: 1;
     top: $block-border-width;
     bottom: $block-border-width;
-    width: $block-padding-width;
+    width: 0;
     border-radius: $block-border-radius;
   }
 


zachmargolis:
Should we also bump the version in this PR, or a separate one?

nickttng:
@zachmargolis We should bump the version. After merge, do you want to handle the npm patch release and take it from there for yours?

zachmargolis:
@nickttng I don't think I have access to npm for this package, maybe that's a @jmhooper thing? 😬

nickttng:
We will bring this up in PR review 👍

PR #135, opened 2020-02-25
Documentation: Update package.json to correct the landing page info

nickttng:
**Why:** On the [design.login.gov](https://design.login.gov/) landing page, the `2.1.6` release number doesn't accurately reflect the [latest npm package release](https://www.npmjs.com/package/identity-style-guide), which is `2.2.1`, potentially confusing team members. 

**How:** 
- Updated the package.json file

## Before
ds-before-2 1 6

## Proposed 
ds-before-2 2 1

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "identity-style-guide",
-  "version": "2.1.6",
+  "version": "2.2.1",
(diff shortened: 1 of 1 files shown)


zachmargolis:
It's neat that the site pulls from the `package.json` and we don't have to update in two places

PR #142, opened 2020-06-08
Run `npm audit fix` to update packages

zachmargolis:
I ran `npm audit fix` so 🤷‍♂️ this might fix some vulnerabilities?

There are 4 left for manual review, but they're dependencies of `gulp` and `get-port-cli` which we're already on the latest versions of

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "serve": "^10.1.2",
     "stylelint": "^9.10.1",
     "stylelint-scss": "^3.9.2",
-    "uswds": "^2.0.3",
+    "uswds": "^2.7.0",
     "uswds-gulp": "github:uswds/uswds-gulp",
     "vinyl-buffer": "^1.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -67,7 +67,7 @@
     "serve": "^10.1.2",
     "stylelint": "^9.10.1",
     "stylelint-scss": "^3.9.2",
-    "uswds": "^2.0.3",
+    "uswds": "^2.7.0",
```
zachmargolis, line 70: This looks like a big change, should we be worried?

nickttng: @zachmargolis Yeah, we should check what has been updated between these two versions - I have a Jira ticket to research this update.

juliaelman:
@zachmargolis would you be open to us closing this PR and make it a part of the USWDS upgrade work?

zachmargolis:
@juliaelman yup! I'll close it, no worries

PR #143, opened 2020-06-19
Small fix login.gov > Login.gov

juliaelman:


zachmargolis:
Federalist build error looks unrelated? Hopefully? Maybe push an empty commit to kick it off again?
> 2020-06-19 15:32:59 ERROR [run-federalist-script] Error: Cannot find module 'uswds-gulp/config/browsers'

juliaelman:
@zachmargolis it looks like there was a change to the gulp build for USWDS in `gulp-uswds` that caused this issue: https://github.com/uswds/uswds-gulp/commit/37bb3ad97a20bcee22bca0f8b528bef1820c85f3

[This seemed to fix the issue](https://github.com/18F/identity-style-guide/pull/143/commits/538bc063376eb2db3532433055f970b8d0d53e0b), but I am not sure how it will effect :all-the-things: 🤔

.browserslistrc
```diff
@@ -0,0 +1,5 @@
+# Supported browsers
+> 2%
+last 2 versions
+IE 11
+not dead
```
zachmargolis, line 5: This is a great file! Never heard of this standard before!

PR #144, opened 2020-06-30
Update`#login-design` Slack channel to `#login-ux`

nickttng:
**Why:** In the slide deck section, the `#login-design` reference should be updated to `#login-ux` to reflect the latest name (and in my not humble opinion, the best channel in all of TTS Slack)


juliaelman:
@nickttng looks like there is an issue with our Federalist build 🤔 maybe try an empty commit to see if it fixes this?

PR #152, opened 2020-08-12
Add DAP tracking

slj:
**Why**: LG-3256
**How**: dap.html partial
**How**: Also upgraded ruby version to 2.6


slj:
https://cm-jira.usa.gov/browse/LG-3256

PR #153, opened 2020-08-12
Add google analytics

slj:
**Why**: LG-3176
**How**: added GA partial


slj:
https://cm-jira.usa.gov/browse/LG-3176

PR #154, opened 2020-08-24
Update Spinner documentation to recommend appropriate pixel density

aduth:
**Why**: Browsers which support srcset can choose amongst given options the most appropriate, including for multiple pixel densities. Since `@2x` is tailored to 2x pixel density, it should ideally only be rendered when the display device has >1 pixel density. Currently, any browser which supports srcset will always show 2x, even if their display is not 2x.

Reference: https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Resolution_switching_Same_size_different_resolutions

Preview URL: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-spinner-pixel-density/components/spinner/

**Screenshots:**

`1x` display|`2x` display
---|---
1x display|2x display

### DIFF
--- a/docs/_components/spinner.md
+++ b/docs/_components/spinner.md
@@ -13,7 +13,7 @@ Spinners can be helpful to signal to a user that the current process may require
 {% capture example %}
 <div class="lg-spinner">
   <div>
-    <img src="{{ site.baseurl }}/assets/img/spinner.gif" srcset="{{ site.baseurl }}/assets/img/spinner@2x.gif" width="144" height="144" class="text-middle" alt="Spinner" />
+    <img src="{{ site.baseurl }}/assets/img/spinner.gif" srcset="{{ site.baseurl }}/assets/img/spinner.gif, {{ site.baseurl }}/assets/img/spinner@2x.gif 2x" width="144" height="144" class="text-middle" alt="Spinner" />
   </div>
 </div>
 {% endcapture %}
@@ -27,7 +27,7 @@ We may want to start with a hidden spinner and then remove it with a button clic
 <button class="usa-button__lg-invokeSpinner">Hide Button and trigger a spinner</button>
 <div class="lg-spinner lg-spinner--hidden" id="spinner-id">
   <div>
-    <img src="{{ site.baseurl }}/assets/img/spinner.gif" srcset="{{ site.baseurl }}/assets/img/spinner@2x.gif" width="144" height="144" class="text-middle" alt="Spinner" />
+    <img src="{{ site.baseurl }}/assets/img/spinner.gif" srcset="{{ site.baseurl }}/assets/img/spinner.gif, {{ site.baseurl }}/assets/img/spinner@2x.gif 2x" width="144" height="144" class="text-middle" alt="Spinner" />
   </div>
 </div>
 {% endcapture %}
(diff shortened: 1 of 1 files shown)


docs/_components/spinner.md
```diff
@@ -13,7 +13,7 @@ Spinners can be helpful to signal to a user that the current process may require
 {% capture example %}
 <div class="lg-spinner">
   <div>
-    <img src="{{ site.baseurl }}/assets/img/spinner.gif" srcset="{{ site.baseurl }}/assets/img/spinner@2x.gif" width="144" height="144" class="text-middle" alt="Spinner" />
+    <img src="{{ site.baseurl }}/assets/img/spinner.gif" srcset="{{ site.baseurl }}/assets/img/spinner.gif, {{ site.baseurl }}/assets/img/spinner@2x.gif 2x" width="144" height="144" class="text-middle" alt="Spinner" />
```
zachmargolis, line 16: maybe easier to read on multiple lines?

```html
    <img src="{{ site.baseurl }}/assets/img/spinner.gif"
         srcset="{{ site.baseurl }}/assets/img/spinner.gif,
                 {{ site.baseurl }}/assets/img/spinner@2x.gif 2x"
         width="144"
         height="144"
         class="text-middle"
         alt="Spinner" />
```

aduth: > maybe easier to read on multiple lines?

Yep, I agree. It definitely reads better. One possible point of confusion is how to align the subsequent lines of the attribute value. In this case, it seems straightforward enough to use spaces to vertically align it with the start of the previous line. If a project uses tabs instead of spaces, it might not be so clear. I think in that case, I might simply recommend to tab one extra level than the preceding line.

docs/_components/spinner.md
```diff
@@ -13,7 +13,15 @@ Spinners can be helpful to signal to a user that the current process may require
...
   <div>
-    <img src="{{ site.baseurl }}/assets/img/spinner.gif" srcset="{{ site.baseurl }}/assets/img/spinner@2x.gif" width="144" height="144" class="text-middle" alt="Spinner" />
+    <img
+      src="{{ site.baseurl }}/assets/img/spinner.gif"
+      srcset="{{ site.baseurl }}/assets/img/spinner.gif,
+              {{ site.baseurl }}/assets/img/spinner@2x.gif 2x"
+      width="144"
+      height="144"
+      class="text-middle"
+      alt="Spinner"
```
aduth, line 23: Not really relevant to the intent of this pull request, but in the course of applying the changes for f7591b6, it struck me whether or not "Spinner" is a good alternative text.

- Is a spinner [decorative](https://www.w3.org/WAI/tutorials/images/decorative/)?
- Or is this a [live region](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-live) which is [busy](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-busy)?
   - https://www.smashingmagazine.com/2015/04/its-alive-apps-that-feed-back-accessibly/
   - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions
- Or maybe "Spinner" is a pragmatic solution that will be reasonably well-understood?

zachmargolis: I think that it's decorative, but we also have an entirely separate page on our accessibility policy and alt text, so I wouldn't worry too much here: https://design.login.gov/accessibility/policies/

PR #157, opened 2020-09-28
Replace usa-input--error-message with usa-error-message (LG-3288, LG-3277)

aduth:
**Why**: For improved alignment with USWDS, to reduce the likelihood of inconsistent error styling.

Relevant USWDS documentation: https://designsystem.digital.gov/components/form-controls/#text-input

Toward LG-3288, this backports the following revisions from IDP:

https://github.com/18F/identity-idp/blob/73c3c13/app/assets/stylesheets/components/_typography.scss#L51-L57

**Open questions:**

- USWDS includes more styling for error messages than what had existed previously with `usa-input--error-message` ([padding, display, font weight](https://github.com/uswds/uswds/blob/22a9eb278196340c3c8909ff5dbeef4ff39756fd/src/stylesheets/elements/form-controls/_global.scss#L54-L59)). Are all of those stylings agreeable, or should we revert back to the original styling? For example, overriding font weight back to `normal`.
- As noted in the associated tickets, USWDS uses the `error-dark` color token for error messages. Based on how errors are styled throughout the IDP, I've opted to keep this as the existing `error` color token in the styling overrides.
- The Login Design System also includes an error message variant including an icon, styled using the `usa-input--error-message-with-icon` class. This pull request does not currently propose any changes to this class, though in considering this as a [BEM modifier](http://getbem.com/naming/) of the error message class, it may be appropriate to update this accordingly to `usa-error-message--with-icon`.

**Screenshots:**

Before|After
---|---
before|after

### DIFF
--- a/docs/_components/form-fields.md
+++ b/docs/_components/form-fields.md
@@ -373,7 +373,7 @@ For radio buttons and checkboxes, simply add an error message directly after the
     </li>
   </ul>
 </fieldset>
-<span class="usa-error-message usa-input--error-message-with-icon" role="alert">Error message text.</span>
+<span class="usa-error-message usa-error-message--with-icon" role="alert">Error message text.</span>
 {% endcapture %}
 {% include helpers/code-example.html code=example %}
     </div>
@@ -405,7 +405,7 @@ For radio buttons and checkboxes, simply add an error message directly after the
     </li>
   </ul>
 </fieldset>
-<span class="usa-error-message usa-input--error-message-with-icon" role="alert">Error message text.</span>
+<span class="usa-error-message usa-error-message--with-icon" role="alert">Error message text.</span>
 {% endcapture %}
 {% include helpers/code-example.html code=example %}
     </div>
@@ -447,7 +447,7 @@ For radio buttons and checkboxes, simply add an error message directly after the
     </li>
   </ul>
 </fieldset>
-<span class="usa-error-message usa-input--error-message-with-icon" role="alert">Error message text.</span>
+<span class="usa-error-message usa-error-message--with-icon" role="alert">Error message text.</span>
 {% endcapture %}
 {% include helpers/code-example.html code=example %}
     </div>
@@ -479,7 +479,7 @@ For radio buttons and checkboxes, simply add an error message directly after the
     </li>
   </ul>
 </fieldset>
-<span class="usa-error-message usa-input--error-message-with-icon" role="alert">Error message text.</span>
+<span class="usa-error-message usa-error-message--with-icon" role="alert">Error message text.</span>
 {% endcapture %}
 {% include helpers/code-example.html code=example %}
     </div>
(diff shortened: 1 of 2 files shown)


aduth:
> * The Login Design System also includes an error message variant including an icon, styled using the `usa-input--error-message-with-icon` class. This pull request does not currently propose any changes to this class, though in considering this as a [BEM modifier](http://getbem.com/naming/) of the error message class, it may be appropriate to update this accordingly to `usa-error-message--with-icon`.

To expand on this, part of the reason I second-guess making this change in the first place is because the previous `usa-input--error-message` class was named as though it were a modifier of some `usa-input` BEM block. With these revisions, `usa-error-message` becomes its own proper block, leaving open the question of whether the "with icon" continues to be a modifier of `usa-input` or of `usa-error-message`.

juliaelman:
Talked with @aduth and had a discussion around how this *might* effect the layout (e.g. cause shifting at various [break points](https://github.com/18F/identity-style-guide/blob/master/src/scss/uswds-theme/_utilities.scss#L33)) may occur. Approved once that has gotten his 👍

juliaelman (approved):

aduth:
Assessing potential impact:

- **Vertical space:** Because `usa-error-message` applies a `0.25rem` padding, surrounding content may be pushed up or down by this amount. In most cases, this should not have a noticeable impact as long as the flow of content on the page can adapt to occupying more height. Notably, the exception here is if the message would be shown in a container of a fixed height. In practice, it does not appear that `usa-input--error-message` is used in any product, and as such would not be expected to regress in existing usage.
- **Block display:** Because `usa-error-message` applies `display: block`, error messages would not be rendered inline and instead on their own line. This is consistent with USWDS guidance, and in most cases has only a positive result. For example, [the "Dates" example of Form validation errors in the live Login Design System website](https://design.login.gov/components/form-fields/#form-validation-errors) has the error message and hint running together without this styling, and would become clearer after these changes.

Before (`usa-input--error-message`)|After (`usa-error-message`)
---|---
before|after

Again, since `usa-input--error-message` is not currently used in any product, there would not be any regressions on existing usage.

aduth:
> * The Login Design System also includes an error message variant including an icon, styled using the `usa-input--error-message-with-icon` class. This pull request does not currently propose any changes to this class, though in considering this as a [BEM modifier](http://getbem.com/naming/) of the error message class, it may be appropriate to update this accordingly to `usa-error-message--with-icon`.

As proposed, in 5ce4aab I opted to rename `usa-input--error-message-with-icon` to `usa-error-message--with-icon`. Since "with icon" is intended to a variant of the error message, BEM conventions would dictate that it be named as such, and not as a modifier of `usa-input`.

Similar to as noted in my previous comment, there is currently no usage of `usa-input--error-message-with-icon` in any product, and as such it would not be expected that this change should cause any regressions. That being said, strictly-speaking it is a breaking change, and if we'd document changes in release notes, it should be mentioned.

PR #158, opened 2020-10-13
Add basic visual regression testing

aduth:
**Why**: As a contributor to Login Design System, I want to compare visual impact of my changes on documented components prior to merge, so that I can have an increased sense of confidence that my changes will not introduce unexpected regressions.

Notably, this is being considered to help facilitate an upgrade of USWDS to the latest version, which may have unanticipated cascading effects on Login Design System.

**Implementation Notes:**

This builds upon existing testing infrastructure, leveraging support for Puppeteer, Jest, and CircleCI artifacts ([source](https://github.com/18F/identity-style-guide/blob/master/.circleci/config.yml#L36-L40), [documentation](https://circleci.com/docs/2.0/artifacts/)) to generate and store screenshot differences between the local working copy and the live documentation site. The missing piece introduced here is an image comparison library, of which [pixelmatch](https://www.npmjs.com/package/pixelmatch) was selected as an optimal candidate.

**Alternatives Considered:**

- [`jest-image-snapshot`](https://github.com/americanexpress/jest-image-snapshot): Good candidate that fits well into model of [Snapshot Testing](https://jestjs.io/docs/en/snapshot-testing). Snapshots require that the files be stored within the repository, which for binary files like images could inflate the download size of the repository over time. Since CircleCI artifact storage is already configured, it seems an ideal preference to reuse for image snapshots. But `jest-image-snapshot` did inspire the use of the same pixelmatch library.
- [Percy.io](https://percy.io/): May provide a much better approval workflow, but is an external (paid) hosted service not currently approved for use. That being said, Percy offers a [Puppeteer integration](https://docs.percy.io/docs/puppeteer) which would appear to be very easy to integrate atop this proposed solution if or when Percy is adopted.

**To do:**

Workflow documentation, pending feedback. Especially...

- Reviewing changes
- Approving acceptable changes

### DIFF
--- a/test/screenshot.test.js
+++ b/test/screenshot.test.js
@@ -81,10 +81,10 @@ function fillImageToSize(image, width, height) {
     for (let x = image.width; x < width; x += 1) {
       // eslint-disable-next-line no-bitwise
       const index = ((resizedImage.width * y) + x) << 2;
-      resizedImage.data[index] = 0;
-      resizedImage.data[index + 1] = 0;
-      resizedImage.data[index + 2] = 0;
-      resizedImage.data[index + 3] = 64;
+      resizedImage.data[index] = 255; // Red
+      resizedImage.data[index + 1] = 255; // Green
+      resizedImage.data[index + 2] = 255; // Blue
+      resizedImage.data[index + 3] = 255; // Alpha (Opacity)
     }
   }
 
--- a/test/screenshot.test.js
+++ b/test/screenshot.test.js
@@ -94,7 +94,7 @@ function fillImageToSize(image, width, height) {
 const testFn = getLatestCommit() === ACKNOWLEDGED_COMMIT ? test.skip : test;
 
 testFn('screenshot visual regression', async () => {
-  const paths = (await getURLsFromSitemap(`${LOCAL_HOST}/sitemap.xml`)).map(getURLPath);
+  const paths = (await getURLsFromSitemap(`${REMOTE_HOST}/sitemap.xml`)).map(getURLPath);
 
   for (const path of paths) {
     const local = await getScreenshot(LOCAL_HOST + path);
--- a/test/screenshot.test.js
+++ b/test/screenshot.test.js
@@ -1,0 +1,130 @@
+/* eslint-disable no-restricted-syntax, no-await-in-loop, no-param-reassign */
+
+const { promises: fsPromises, readFileSync } = require('fs');
+const { join } = require('path');
+const { spawnSync } = require('child_process');
+const assert = require('assert');
+const mkdirp = require('mkdirp');
+const { PNG } = require('pngjs');
+const match = require('pixelmatch');
+const YAML = require('yaml');
+
+const { writeFile } = fsPromises;
+
+const ACKNOWLEDGED_COMMIT = '247b7e6815b358d7962ad2bc21d34dd9122bbe69';
+const LOCAL_HOST = `http://localhost:${process.env.JEST_PORT}`;
+const REMOTE_HOST = 'https://design.login.gov';
+const DIFF_DIRECTORY = join(__dirname, '../tmp/results/screenshot-diff');
+const { url: URL_PREFIX, github_repo_url: GITHUB_REPO_URL } = YAML.parse(
+  readFileSync(join(__dirname, '../_config.yml'), 'utf8'),
+);
+
+async function getURLsFromSitemap(url) {
+  await page.goto(url);
+  return page.$$eval('url loc', locs => locs.map(loc => loc.textContent));
+}
+
+async function stubAnimations() {
+  await page.evaluate(() => {
+    const isGif = img => img.src.endsWith('.gif');
+    function stubGif(img) {
+      img.width = img.width;
+      img.height = img.height;
+      img.src = 'data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==';
+      img.removeAttribute('srcset');
+    }
+
+    [...document.querySelectorAll('img')].filter(isGif).forEach(stubGif);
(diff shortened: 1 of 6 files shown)


aduth:
Example failure can be seen at #159:

1. Click Details

Screen Shot 2020-10-13 at 11 21 24 AM

2. See build failure in CircleCI

image

3. Switch to artifacts table in CircleCI

Screen Shot 2020-10-13 at 11 22 07 AM

4. Browse and view diff artifacts

Screen Shot 2020-10-13 at 11 22 51 AM

---

Local|Remote|Diff
---|---|---
image|image|image

This particular issue seems to be an overflow problem causing the horizontal width of the page to grow after the USWDS upgrade. The diff is not especially useful here, since the screenshots are so drastically different.

test/screenshot.test.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  return resizedImage;
+}
+
+const testFn = getLatestCommit() === ACKNOWLEDGED_COMMIT ? test.skip : test;
+
+testFn('screenshot visual regression', async () => {
+  const paths = (await getURLsFromSitemap(`${LOCAL_HOST}/sitemap.xml`)).map(getURLPath);
```
zachmargolis, line 97: WDYT of using the remote sitemap? because if we add a new page, it'll only exist locally and won't exist remote, so it'll automatically fail right?

test/screenshot.test.js
```diff
@@ -0,0 +1,130 @@
...
+  PNG.bitblt(image, resizedImage, 0, 0, image.width, image.height, 0, 0);
+
+  for (let y = image.height; y < height; y += 1) {
+    for (let x = image.width; x < width; x += 1) {
+      // eslint-disable-next-line no-bitwise
+      const index = ((resizedImage.width * y) + x) << 2;
+      resizedImage.data[index] = 0;
+      resizedImage.data[index + 1] = 0;
+      resizedImage.data[index + 2] = 0;
+      resizedImage.data[index + 3] = 64;
```
zachmargolis, line 87: WDYT of commenting these? I'm taking a guess here
```suggestion
      resizedImage.data[index] = 0; // red
      resizedImage.data[index + 1] = 0; // green
      resizedImage.data[index + 2] = 0; // blue
      resizedImage.data[index + 3] = 64; // alpha
```

zachmargolis (approved):
LGTM! Love how we're using the deployed site instead of checking in large image assets to the repo

For "acknowledgement" I bet we could probably get pretty far with making the visual diff a separate CircleCI job, and not making it required, and making sure we leave some sort of "I'm acknowledging the diff error" comment when we merge?

test/screenshot.test.js
aduth: > WDYT of using the remote sitemap? because if we add a new page, it'll only exist locally and won't exist remote, so it'll automatically fail right?

Yeah, that makes sense to me.

Originally I'd noticed the URLs would be the simple paths when in development mode, which (combined with shorter turnaround time) was the primary reason for using the local URL. I later discovered the pre-test script will build for production, so we need to trim the [prefix URL](https://github.com/18F/identity-style-guide/blob/086d240b4bad78f9e7bc892ee7d907fe9d085a96/_config.yml#L4) anyways. Swapping to the remote URL would be trivial then, and avoid those issues you describe.

aduth:
> For "acknowledgement" I bet we could probably get pretty far with making the visual diff a separate CircleCI job, and not making it required, and making sure we leave some sort of "I'm acknowledging the diff error" comment when we merge?

I like the idea, also in how it would show as a separate check entry, and in (I assume?) being a simple way to parallelize these long-running tests as being independent from the rest. As far as how to actually implement this between GitHub Checks and CircleCI conditions, is your thinking to allow this to show as a failure, but still be merge-able from within GitHub? Or were you thinking something different?

It does remind me of another alternative I'd had in mind, which was on the same line of thinking of turning this less into an assertive failure, and more into an "FYI, these pages changed, in case you weren't expecting it". I'd even thought of making some sort of standalone GitHub Action which wouldn't actually integrate as a status check, but would instead simply leave a comment on the pull request with any diffs. In this way, it's hard to miss, but at the same time isn't treated as a failure per-se. That being said, it seemed it would involve a fair bit more effort to go down this route.

test/screenshot.test.js
aduth: > WDYT of commenting these? I'm taking a guess here

Updated to add comments in ed514d3.

test/screenshot.test.js
aduth: > WDYT of using the remote sitemap? because if we add a new page, it'll only exist locally and won't exist remote, so it'll automatically fail right?

Updated in 1cddbd0.

zachmargolis:
>  As far as how to actually implement this between GitHub Checks and CircleCI conditions, is your thinking to allow this to show as a failure, but still be merge-able from within GitHub?

Correct, I was thinking we'd make it a separate job. Since we can toggle required/optional per-job in GH, I am thinking we'd leave this particular job as optional and the rest of the CI as required

And yes, seems like a great candidate for parallelization too

aduth:
I've updated the CircleCI configuration to break down a few independent jobs in general alignment with the IDP (9397ea2, 1274cc2, 49d1a19, 3b90a27)

I also added some documentation in 0bbb5a3 describing what I assume to be the expected workflow.

I don't have access to the repository settings, so I can't configure this check as being optional. @zachmargolis Would you be able to take care of this? It might also be something which can only be done after merge, to allow for the new checks to take effect.

zachmargolis:
> I don't have access to the repository settings, so I can't configure this check as being optional. @zachmargolis Would you be able to take care of this? It might also be something which can only be done after merge, to allow for the new checks to take effect.

New checks default to optional, so we're all good here 👍

PR #160, opened 2020-10-15
Toggle hidden content visible for visual regression tests

aduth:
**Why**: Check for regressions of content which may be collapsed (hidden) by default.


juliaelman (approved):
🚀

PR #161, opened 2020-10-15
Update out-of-sync Gemfile

aduth:
Introduced in #147

**Why**: Out of sync with Gemfile

Adds new lint test to verify no uncommitted changes in local working copy after running `bundle install` and `npm install`.

**Why**: Prevent uncommitted changes on install


zachmargolis (approved):
LGTM

PR #163, opened 2020-10-19
Only check diffs of lockfiles in lint task

aduth:
Couple small improvements to lockfile validation implemented in #161:

- 2e48906 Only check diffs of lockfiles in lint task
  - **Why**: It may be common to run `npm run lint` before committing changes unrelated to lockfiles, which should not fail due to the fact that there are uncommitted changes.
- fe1c664 Ignore scripts in package-lock.json validation npm install
  - **Why**: Faster. Especially since this project runs `bundle install` as a postinstall script, which in lockfile validation is redundant with validate-gemfile-lock. Still satisfies purpose of lockfile validation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ validate-gemfile-lock: Gemfile Gemfile.lock
 
 validate-package-lock: package.json package-lock.json
 	@echo "Validating package-lock.json..."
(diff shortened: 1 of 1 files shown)


zachmargolis (approved):
LGTM

PR #164, opened 2020-10-20
Fix basic styles setting assignment

aduth:
**Why**: The design system assigns `$theme-global-styles-basic: true;` with the intention of assigning basic styles for common base elements, but this does not apply, because the `$theme-global-styles-basic` setting does not exist. It was replaced with three separate variables as part of the USWDS 2.0.0 beta release process:

>- Replaced the `$theme-global-styles-basic` variable with three explicit global styling vars — each set to `false` by default.
>   - `$theme-global-paragraph-styles` applies styling to `p` elements
>   - `$theme-global-link-styles` applies styling to `a` elements
>   - `$theme-global-content-styles` applies styling to `p`, `a`, `h[1-6]`, `ol`, `ul`, and `table` elements

https://designsystem.digital.gov/about/releases/#version-200-beta-6

~This is not quite ready for review, because it causes regressions discovered in visual regression tests (which, as an aside, serves conveniently for demonstration of visual regression tests).~

**Edit:** See comments below. As discussed, the intention is to keep the global styles as not applied by default.

### DIFF
--- a/src/scss/uswds-theme/_typography.scss
+++ b/src/scss/uswds-theme/_typography.scss
@@ -53,16 +53,20 @@ $theme-root-font-size: 16px;
 
 /*
 ----------------------------------------
-Basic styles
+Global styles
 ----------------------------------------
-Sets the default typography styles for
-paragraph text and links
-
-Accepts true or false
+Adds basic styling for the following
+unclassed elements:
+- paragraph: paragraph text
+- link: links
+- content: paragraph text, links,
+  headings, lists, and tables
 ----------------------------------------
 */
 
-$theme-global-styles-basic: true;
+$theme-global-paragraph-styles: true;
+$theme-global-link-styles: true;
+$theme-global-content-styles: true;
 
 /*
 ----------------------------------------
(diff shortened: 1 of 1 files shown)


aduth:
On further discussion, it was decided to not apply these styles globally, but instead to add styles (`usa-link`, etc) explictly.

aduth:
> On further discussion, it was decided to not apply these styles globally, but instead to add styles (`usa-link`, etc) explicitly.
> 
> This pull request will need to be updated to _remove_ the code, and optionally to add `usa-link` and other "base" styles to the design system documentation pages.

To clearly communicate that this is intended to be explicitly set, I opted to keep the variables, but assign the variables as `false` in 211795b.

I'll plan to address adding `usa-link` to all links separately.

zachmargolis (approved):
LGTM

aduth:
> I'll plan to address adding `usa-link` to all links separately.

I split this into its own discovery ticket (LG-3768). Depending on the complications described there, we may want to revisit this discussion. In any case, these changes at least better reflect the current state of configuration.

PR #165, opened 2020-11-16
LG-3757: Create bundler-ready module entrypoint

aduth:
**Why**: As a developer using the login.gov Design System, I expect that I can use identity-style-guide as a proper NPM package, so that I can deduplicate shared dependencies (see LG-3716) and take advantage of dead code elimination to reduce bundle size by omitting unused components.

**Changelog:**

>- The published NPM package can now be imported into a project. Bundlers that support dead-code elimination (Webpack, etc.) can automatically remove unused code to optimize bundle size. See [Usage guide](https://design.login.gov/usage/) for more information. Existing usage of JavaScript code from `assets/` is unaffected by these changes.

**Usage Documentation:** https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-lg-3716-package-entry/usage/#use-as-a-javascript-package

**History:**

#148 added a main to the package's package.json, which at least allows for a developer to import the package. However, because this points to a precompiled distributable asset, dependencies cannot be deduplicated.

**Impact:**

Preliminary testing in `identity-idp` reveals bundle size reduction of 88% gzipped (451kb to 56.3kb), largely attributed to elimination of unused `zxcvbn` dependency.

<table><thead><tr><th>Before</th><th>After</th></tr></thead><tbody><tr><td><pre><code>js/application-117899d0f4756bf36d57.js         1 MiB    2, 3  [emitted] [immutable]  [big]  application
js/application-117899d0f4756bf36d57.js.br    402 KiB          [emitted]              [big]
js/application-117899d0f4756bf36d57.js.gz    451 KiB          [emitted]              [big]</code></pre></td><td><pre><code>js/application-646a51c2eba309441c63.js       203 KiB    2, 3  [emitted] [immutable]         application
js/application-646a51c2eba309441c63.js.br   47.7 KiB          [emitted]
js/application-646a51c2eba309441c63.js.gz   56.3 KiB          [emitted]</code></pre></td></tr></tbody></table>


aduth:
The Snyk advisory is arguably not worth concerning over, but it can't be resolved without upgrading USWDS, incidentally resolved by #159.

src/js/index.js
```diff
@@ -0,0 +1,13 @@
+export {
```
zachmargolis, line 1: Why do we have both this file and `./components`? Seems like it would be easy to accidentally forget updating one or the other, can we combine them?

zachmargolis (approved):
LGTM, one question

aduth:
Also, should have mentioned earlier that the "visual-regression" failure is expected: It's catching the new content added to the "Usage" page. 😅 

At some point I'd like to revisit this to ideally report as a comment and not as a build failure.

src/js/index.js
aduth: Yeah, it's a good point. The intention was to share the set of exports from `components/index.js` between the package entrypoint and the browser "auto" bundle, but considering that the package entrypoint is already a side-effect-free set of exports, the auto bundle could just read from it.

I'll plan to just import each component from this file instead.

Alternatively, we could use [`export * from './components';`](https://github.com/tc39/proposal-export-ns-from). I was under the wrong impression that the proposal hadn't progressed very far, but looks like it's readily available now.

To the general concern of forgetting to keep things in sync, this was the idea with the tests in `test/components/index.test.js`, since we'll have more USWDS components arriving with 2.9.0 (#159) and I'd worried I'd forget to include them 😄  though in that case it's less within our control, since those components are defined in USWDS>

aduth: Updated in 9672158 (using `export * from` syntax).

aduth:
Rebased after merging #159:

- Snyk advisory should hopefully clear up.
- Added components introduced in recent versions of USWDS (db52eee).
- For CHANGELOG.md, reconciled as "New Features" of 3.0.0 major version.

aduth:
Visual regression shows small adjustment of padding in the header. It looks that this may have been some flakiness in the rebuild from #159, since it looks more correct in the "changes", and the "changes" are reflected even in what I see in the `master` branch run locally.

PR #166, opened 2020-11-17
Implement dropdown button as component

aduth:
Related: #165

The [dropdown button](https://design.login.gov/components/buttons/#dropdown) JavaScript implementation was not previously implemented as a USWDS-conventional component. Because it contained side-effects to initialize a dropdown button on a page, it was excluded from the package in #165, and only available in the `main.js` precompiled JavaScript bundle. This pull request refactors the code to bring it in line with other components.

Incidentally, this also resolves an issue where only one dropdown button could exist on a page.

**Future Work:**

- Button should assign `aria-controls` to complement existing `aria-expanded` ([documentation](https://www.w3.org/WAI/PF/aria-1.1/states_and_properties#aria-expanded)).
   - >If the element with the `aria-expanded` attribute controls the expansion of another grouping container that is not 'owned by' the element, the author **SHOULD** reference the container by using the `aria-controls` attribute.
- Consider deprecating, or at least finding a way to maximize reuse of similar accordion behaviors, which already handle most of `aria-expanded` and `aria-controls` behavior.
- Consolidate to eliminate separation between mobile and desktop markup (impacted dashboard only appears to use desktop markup anyways)
   - In the process, consider renaming classes to follow standard conventions (`usa-` or `lg-` prefixed)

**Impact:**

The button only appears to be used in the dashboard, in a view which may in-fact not be used? ([source](https://github.com/18F/identity-dashboard/blob/master/app/views/manage_users/_add_user_options_dropdown.html.erb))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,5 @@
+## 2.2.4
+
+### Bug Fixes
+
+- Fixed support for multiple dropdown buttons on a page.
--- a/src/js/components/dropdownButton.js
+++ b/src/js/components/dropdownButton.js
@@ -1,34 +1,19 @@
-document.addEventListener('DOMContentLoaded', () => {
-  const mobileLink = document.querySelector('.dropdown-mobile-toggle > button');
-  const mobileDropdown = document.querySelector('.mobile-dropdown');
-  const desktopLink = document.querySelector('.dropdown-desktop-toggle > button');
-  const desktopDropdown = document.querySelector('.desktop-dropdown');
+const behavior = require('uswds/src/js/utils/behavior');
 
-  function addListenerMulti(el, s, fn) {
-    s.split(' ').forEach((e) => el.addEventListener(e, fn, false));
-  }
+function createDropdownPicker(contentSelector) {
+  return function onDropdownButtonClickOrKeyPress(event) {
+    const button = event.target;
+    const content = button.parentNode.querySelector(contentSelector);
+    const isExpanded = !content.classList.toggle('display-none');
+    button.setAttribute('aria-expanded', isExpanded.toString());
+  };
+}
 
-  function toggleAriaExpanded(element) {
-    if (element.getAttribute('aria-expanded') === 'true') {
-      element.setAttribute('aria-expanded', 'false');
-    } else {
-      element.setAttribute('aria-expanded', 'true');
-    }
-  }
-
-  function dropdownPicker(trigger, dropdown) {
-    addListenerMulti(trigger, 'click keypress', function (event) {
-      const eventType = event.type;
-
-      event.preventDefault();
-      if (eventType === 'click' || (eventType === 'keypress' && event.which === 13)) {
-        this.parentNode.classList.toggle('focused');
-        dropdown.classList.toggle('display-none');
-        toggleAriaExpanded(this);
-      }
-    });
-  }
-
-  if (desktopLink) dropdownPicker(desktopLink, desktopDropdown);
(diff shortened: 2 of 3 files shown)


src/js/components/dropdownButton.js
```diff
@@ -1,34 +1,19 @@
-document.addEventListener('DOMContentLoaded', () => {
```
zachmargolis, line 1: Without this event listener, does calling code need to manually activate this behavior now? I tried to look at how `behavior()` in USWDS was implemented and didn't see anything that listened for DOM ready

src/js/components/dropdownButton.js
```diff
@@ -1,34 +1,19 @@
...
+const behavior = require('uswds/src/js/utils/behavior');
 
-  function addListenerMulti(el, s, fn) {
-    s.split(' ').forEach((e) => el.addEventListener(e, fn, false));
-  }
+function createDropdownPicker(contentSelector) {
+  return function onDropdownButtonClickOrKeyPress(event) {
+    const button = event.target;
+    const content = button.parentNode.querySelector(contentSelector);
+    const isExpanded = !content.classList.toggle('display-none');
```
zachmargolis, line 7: Does USWDS polyfill `classList` already for us?

CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+## 2.2.4
```
zachmargolis, line 1: 😬 this is gonna conflict with https://github.com/18F/identity-style-guide/pull/159/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

zachmargolis (approved):
LGTM

CHANGELOG.md
aduth: > 😬 this is gonna conflict with https://github.com/18F/identity-style-guide/pull/159/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

Yeah, there's going to be a lot of rebase surgery to be done between all the open pull requests 😄

src/js/components/dropdownButton.js
aduth: > Does USWDS polyfill `classList` already for us?

Yeah:

https://github.com/uswds/uswds/blob/92ae6f8/src/js/polyfills/index.js#L1-L2

src/js/components/dropdownButton.js
aduth: > Without this event listener, does calling code need to manually activate this behavior now? I tried to look at how `behavior()` in USWDS was implemented and didn't see anything that listened for DOM ready

It's a bit indirect, but no, a consumer of the existing `dist/assets/js/main.js` won't need to change anything to initialize the dropdown button. It'll still happen on `DOMContentLoaded` automatically.

We handle this by iterating through `uswds.components` on `domready` to initialize each component, which has effectively the same result as what existed previously:

https://github.com/18F/identity-style-guide/blob/d900161552f4b2990a07f5f1009c1a776804ec36/src/js/main.js#L21-L28

With #165, a consumer would need to initialize it themselves if they use the component through the new package entrypoint, but that's intended, and also [documented](https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-lg-3716-package-entry/usage/#use-as-a-javascript-package).

zachmargolis: Ah perfect, thanks for clarifying

PR #168, opened 2020-11-18
Fix incorrect footer reference on layout page

orenyk:


juliaelman (approved):
🚀

zachmargolis (approved):
LGTM, the visual regression is from the copy change, which is intentional

PR #170, opened 2020-11-30
Update Rails usage docs to include Sprockets config

orenyk:
Sprockets needs to be explicitly told about the images to precompile if they're not referenced in the SASS/JS.

Note that this is just the workaround that I came up with for my use case, but if there's a better way please let me know 😄.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -97,6 +97,52 @@ $image-path: 'img';
 @import 'identity-style-guide/dist/assets/scss/styles';
 ```
 
+If you're using Sprockets and precompiling assets you'll need to update your
+Sprockets manifest to include the images and fonts for production:
+
+```js
+// app/assets/config/manifest.js
+
+//= link_tree ../../../node_modules/identity-style-guide/dist/assets/img
+//= link_tree ../../../node_modules/identity-style-guide/dist/assets/fonts
+```
+
+Unfortunately, this results in the assets being saved under paths that include
+everything after the `node_modules` directory, so a helper method is useful in
+cleaning up the views:
+
+```ruby
+# app/helpers/assets_helper.rb
+
+module AssetsHelper
+  def login_design_asset_path(path)
+    "identity-style-guide/dist/assets/#{path}"
+  end
+end
+```
+
+```erb
+# app/views/foo.html.erb
+
+<%= image_tag login_design_asset_path('img/us_flag_small.png'), ... %>
+
+# app/views/layouts/application.html.erb
+
+<head>
+  ...
+  <%= favicon_link_tag identity_asset_path('img/favicons/favicon.ico') %>
+</head>
+```
+
+Finally, if you're using Webpacker you'll need to reference the JS files in the
+default pack:
+
+```js
+// app/javascript/packs/application.js
+
+require("identity-style-guide/dist/assets/js/main")
+```
+
 ### Use as a JavaScript package
 
 If you're already using a JavaScript bundler in your project, you can import specific component implementations from the `identity-style-guide` package. Most modern bundlers that support dead-code elimination will automatically optimize the bundle size to include only the code necessary in your project.


aduth:
Hm, I was under the impression it was the previous snippet of text in the documentation which is key to making the assets available. Namely, adding `node_modules` to `Rails.application.config.assets.paths`:

```rb
Rails.application.config.assets.paths << Rails.root.join('node_modules')
```

This is how it's done in the IdP:

https://github.com/18F/identity-idp/blob/f8f0515/config/initializers/assets.rb#L7

And then referencing assets using the URL asset helper methods, e.g.:

https://github.com/18F/identity-idp/blob/f8f0515/app/views/layouts/base.html.erb#L26-L28

I see there's some equivalent code in [`identity-dashboard/config/initializers/assets.rb`](https://github.com/18F/identity-dashboard/blob/master/config/initializers/assets.rb). Does it not work when you reference assets similarly, without the additional configuration you're proposing to be included here?

orenyk:
@aduth I will play around and let you know, thanks for the follow-up!

docs/usage.md
```diff
@@ -97,6 +97,42 @@ $image-path: 'img';
...
+Unfortunately, this results in the assets being saved under paths that include
+everything after the `node_modules` directory, so a helper method is useful in
+cleaning up the views:
+
+```ruby
+# app/helpers/assets_helper.rb
+
+module AssetsHelper
+  def identity_asset_path(path)
+    "identity-styly-guide/dist/assets/#{path}"
```
aduth, line 118: Typo:

```suggestion
    "identity-style-guide/dist/assets/#{path}"
```

docs/usage.md
```diff
@@ -97,6 +97,42 @@ $image-path: 'img';
 @import 'identity-style-guide/dist/assets/scss/styles';
 ```
 
+If you're using Sprockets and precompiling assets you'll need to update your
+Sprockets manifest to include the images for production:
+
+```js
+// app/assets/config/manifest.js
+
+//= link_tree ../../../node_modules/identity-style-guide/dist/assets/img
```
aduth, line 106: Since we may want to reference more than just images, should this be....

```suggestion
//= link_tree ../../../node_modules/identity-style-guide/dist/assets
```

Thinking specifically how we're referencing fonts in `identity-idp`.

If we update, we'll probably want to update the preceding text as well, since it mentions "images for production". "assets for production" might be more applicable in this case.

docs/usage.md
```diff
@@ -97,6 +97,42 @@ $image-path: 'img';
...
+
+Unfortunately, this results in the assets being saved under paths that include
+everything after the `node_modules` directory, so a helper method is useful in
+cleaning up the views:
+
+```ruby
+# app/helpers/assets_helper.rb
+
+module AssetsHelper
+  def identity_asset_path(path)
```
aduth, line 117: "identity" seems a little broad to reference the design system assets specifically, though I'm not sure I have a better alternative to offer. `lgds_asset_path` works but the abbreviation may not be obvious to everyone. `login_design_system_asset_path` or variations thereof can work and are descriptive, but exceedingly so.

aduth (approved):
@orenyk and I worked through this together in Slack this morning, and he helped illustrate the problem to me. Specifically, I was confused in thinking these assets were readily available without additional configuration or manifest reference based on a few examples from the `identity-idp` repository (fonts specifically). In those cases, they were only precompiled because they were picked up as referenced by stylesheets.

docs/usage.md
orenyk: oof thanks!

docs/usage.md
orenyk: Do you think we should reference the subdirectories explicitly? I'm thinking it might get confusing for Sprockets if it tries to precompile the SCSS and JS files separately from those parts of the asset pipeline.

docs/usage.md
orenyk: Maybe `login_design_asset_path` to keep it from getting too long?

docs/usage.md
orenyk: After looking at the `dist/assets` directory I think we should explicitly reference the fonts and images rather than the whole folder.

aduth: Sounds good 👍  I agree with your point about JS and SCSS.

docs/usage.md
aduth: > Maybe `login_design_asset_path` to keep it from getting too long?

Any of the above would be fine with me. Even keeping it as-is could work, considering it's just an example and could be customized specific to the project where it's implemented.

aduth (approved):

PR #171, opened 2020-12-03
Replace deprecated USWDS variable references with updated equivalent values

aduth:
**Why**: This variable was deprecated in USWDS 2.2.0. Using the updated equivalent puts us in a better position to handle a major version bump of USWDS.

**Expected impact:** There is not expected to be a change in behavior with these revisions, as the deprecated variable was defined as pointing to the new variable value.

**References:**
- https://designsystem.digital.gov/about/releases/#version-220
- https://github.com/uswds/uswds/blob/18cd29e/src/stylesheets/core/_deprecated.scss#L20-L21

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 3.0.1
+
+### Internal
+
+- Replace deprecated USWDS variable references with updated equivalent values.
+
 ## 3.0.0
 
 ### Breaking Changes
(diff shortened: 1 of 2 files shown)


aduth:
Re: Visual regression testing. There appears to be an issue with compiling the site where stylesheets are loaded in a non-deterministic order. Needs more investigation.

zachmargolis (approved):
LGTM

aduth:
Going to go ahead and merge this and address the visual regression tests separately, since it appears to be a general flakiness of the build.

PR #172, opened 2020-12-04
Remove css-mqpacker dependency

aduth:
**Why**: May produce non-deterministic style ordering. Package is deprecated.

See: https://github.com/hail2u/node-css-mqpacker#css-cascading-order

Related: https://github.com/18F/identity-style-guide/pull/171#issuecomment-738078540

**Impact:**

Because this package tried to merge media queries, it should be expected there's a small increase in size of raw built CSS files with its removal. Surprisingly (or not, given how gzip [actually works](https://en.wikipedia.org/wiki/DEFLATE#Duplicate_string_elimination)), the gzip size is _reduced_ after these changes.

&nbsp;|Before|After
---|---|---
raw|328k|340k
gzip|49.4k|47.7k


gulpfile.js
```diff
@@ -116,7 +115,6 @@ gulp.task('build-sass', () => {
       cascade: false,
       grid: true,
     }),
-    mqpacker({ sort: true }),
```
zachmargolis, line 119: is there some other kind of CSS sort we can use instead to get the determinism back?

zachmargolis (approved):
LGTM

gulpfile.js
aduth: > is there some other kind of CSS sort we can use instead to get the determinism back?

Yeah, there's at least [two options](https://github.com/hail2u/node-css-mqpacker#sort) here. To be honest, I didn't spend much time considering it, since I don't really think this package is worth keeping around, all things considered (it's deprecated, maintenance overhead, unpredictability of build-time reordering, actually increasing the size of the gzipped bundle).

zachmargolis: No I meant some sort of other part of the build process instead of `mqpacker` -- not options within `mqpacker`

mitchellhenke (approved):

aduth:
Visual regression tests are a bit perplexing, but lead me to think that USWDS actually relies on this behavior. They'd switched off to another alternative at some point.

https://github.com/uswds/uswds/commit/2f88b5da1f62c5265c96796959638190033893c9

Will do more testing...

**Edit:**

>Note: media query sorting is no longer required as of USWDS 2.5.0. We stopped sorting media queries with csso in USWDS 2.5.1 because it wasn't outputting as expected. While both the minified and unminified CSS files are modestly larger as a result: 268 KB unsorted vs. 259 KB sorted, our testing indicates that once the files are compressed server side with gzip, the unsorted CSS is actually smaller: 36 KB unsorted and gzipped vs. 38 KB sorted and gzipped. As a result, we recommend that teams do not use media query sorting at this time.  

https://github.com/uswds/uswds/blob/develop/README.md#sass-compilation-requirements

aduth:
Debugging results:

- We previously relied on the reordering for the USWDS `.usa-display` styles to take precedent over [our own explicit styling](https://github.com/18F/identity-style-guide/blob/d2d697bb95bd9677e002d06fea1ac7d62c1b2902/src/scss/components/_typography.scss#L83-L85), which no longer happens after removing `css-mqpacker`. To restore the previous behavior, it's merely a matter of removing our custom styling.
- We assign [`$theme-title-font-size`](https://github.com/18F/identity-style-guide/blob/d2d697bb95bd9677e002d06fea1ac7d62c1b2902/src/scss/uswds-theme/_typography.scss#L306) variable which presumably we expected to override the `.usa-display` font size, but this doesn't happen because the variable [was deprecated in 2.0.2](https://github.com/uswds/uswds/blob/b0e7ea3800e5bb5960e0925716cf5363375ee613/src/stylesheets/core/_deprecated.scss#L24-L25) (even preceding this week's version bump) and the heading styles are only applied when configured using the updated variable name ([source](https://github.com/uswds/uswds/blob/b0e7ea3800e5bb5960e0925716cf5363375ee613/src/stylesheets/core/mixins/_typography.scss#L83)).

Summary:

- Display heading should be "restored" after recent commits, but in this case "restored" means to what I assume we expect the display titles to be shown as (`2xl`), which is never what they were actually shown as (actually `3xl`, USWDS' default).

gulpfile.js
aduth: > No I meant some sort of other part of the build process instead of `mqpacker` -- not options within `mqpacker`

Ah. I don't think we want the build to manage this. Instead, we should want to rely on manual ordering of our imports (e.g. [settings](https://github.com/18F/identity-style-guide/blob/d2d697bb95bd9677e002d06fea1ac7d62c1b2902/src/scss/styles.scss#L3-L9), [dependencies](https://github.com/18F/identity-style-guide/blob/d2d697bb95bd9677e002d06fea1ac7d62c1b2902/src/scss/styles.scss#L11), [overrides and extras](https://github.com/18F/identity-style-guide/blob/d2d697bb95bd9677e002d06fea1ac7d62c1b2902/src/scss/styles.scss#L13-L30), [similar within USWDS itself](https://github.com/uswds/uswds/blob/b0e7ea3800e5bb5960e0925716cf5363375ee613/src/stylesheets/uswds.scss#L3-L17)).

Or maybe I'm still not understanding 😅   As far as I can tell, there _was_ an expected determinism by how we order the stylesheets, and the expectation was broken due to how mqpacker works to regroup the output styles.

zachmargolis: Ok! So we did want manual ordering, I think it all makes sense now

PR #173, opened 2020-12-07
Use ESM import/export syntax for dropdown button

aduth:
**Why**: When using the ES module entrypoint, CommonJS "module.exports" and "require" are invalid.

This should only be an issue when using the ES module entrypoint `build/esm/index.js`, and not an issue for CommonJS usage (`build/cjs/index.js`) or the precompiled copy (`dist/assets/js/main.js`).

This was likely missed due to merge timing between #166 and #165.


mitchellhenke (approved):
👍🏼

zachmargolis (approved):
LGTM

PR #177, opened 2021-01-15
Remove "Coming soon!" pages from the design system.

juliaelman:
These components will be added when they are implemented and designed.


zachmargolis (approved):
LGTM

zachmargolis:
Visual regression is expected since we removed pages and items from nav ✔️

PR #178, opened 2021-01-20
Remove accessibility statement from design.login.gov in favor of it being on login.gov

juliaelman:
Login.gov has recently updated our content, look and feel to reflect our growing program. The accessibility content on [design.login.gov](https://design.login.gov/accessibility/) is the same content that exists on [login.gov](https://login.gov/accessibility/). 

A few things to note before reviewing this PR:

1. The same content exists on login.gov, which is the main site for the program. This gives accessibility more prominence and importance to our efforts and the attention it deserves.
2. Accessibility guidance will exist at the component level so that there is specific guidance for those components, rather than overarching guidance.
3. This PR includes a re-direct to the new accessibility page on login.gov in case there are any lingering urls that used to link to this page on design.login.gov.


docs/_data/nav.yml
```diff
@@ -2,7 +2,6 @@ primary:
   - href: /usage/
   - href: /brand/
   - href: /content/
-  - href: /accessibility/
```
zachmargolis, line 5: do we want to leave in some other way for this page to be navigable from the design site?

zachmargolis (approved):
LGTM

docs/_data/nav.yml
aduth: > do we want to leave in some other way for this page to be navigable from the design site?

I think the intention is to remove it completely from the design system website, and that the redirect exists largely to prevent dead links in case it's referenced elsewhere. @juliaelman can confirm though if this aligns to her expectations.

juliaelman: @zachmargolis great question. Talked with Nick. We will be updating the design.login.gov footer to match the footer on login.gov so that the statement is added there and we have a consistency in branding.

docs/accessibility.md
```diff
@@ -1,65 +1,5 @@
...
-
-Our completed accessibility assessment reports can be found below:
-
-[IAL 1 Account creation report](https://docs.google.com/document/d/1wqbVOvNg-7tFW1l4ZxMJrA_VR2gQqNpimlbzxPtvpXg/export?format=pdf)
-
-### Accessibility policies
-
-In addition to completed accessibility reports, Login.gov has compiled a page regarding accessibility policies.
-
-[Login.gov accessibility policies]({% link accessibility/policies.md %})
```
aduth, line 43: Do we want to remove this page as well?

https://design.login.gov/accessibility/policies/

Since we don't link to it anywhere now.

PR #180, opened 2021-01-20
Enable Jekyll redirect plugin

aduth:
**Why**: So that accessibility redirect introduced in #178 does redirect.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -10,6 +10,7 @@ keep_files:
 
 plugins:
   - jekyll-sitemap
+  - jekyll-redirect-from
 
 collections:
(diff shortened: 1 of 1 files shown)


zachmargolis (approved):
LGTM

juliaelman (approved):

aduth:
Visual regression test failure reminds me that these pages are still included in `sitemap.xml` since the two plugins `jekyll-sitemap` and `jekyll-redirect-from` aren't aware of each other and we need to manually [exclude](https://github.com/jekyll/jekyll-sitemap#exclusions).

PR #182, opened 2021-01-26
Link to base USWDS component documentation

aduth:
**Why**: As someone using the login.gov Design System, I want to know when a component is implemented as an extension of the U.S. Web Design System, so that I can reference guidance from the USWDS documentation, and see specific implementation differences contained in the component stylesheet.

(Design has been roughly adapted from Summit Hackathon Figma document)

**Screenshot:**

image


zachmargolis (approved):
LGTM

PR #181, opened 2021-01-20
Remove accessibility policies page

aduth:
As of #178, the "Accessibility" top-level page was removed. This left a subpage "Accessibility Policies" no longer linked to, but it still exists in the site:

https://design.login.gov/accessibility/policies/

The proposed changes here adapt the content of this page to a new "Images" page, to include guidance on how login.gov uses illustration, how those illustrations should often be considered decorative, and accessibility recommendations for marking up those images accordingly.

**Live preview URL:** https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-remove-a11y-policies/images/

**Screenshot:**

images page screenshot

PR #184, opened 2021-01-29
Fix prose sibling heading margin selectors

aduth:
**Why**: It's assumed _(confirmation needed)_ that the intention of these selectors was to match headings within `.usa-prose` following any other content in that prose block. Instead, due to the behavior of `&` in expanding the full selector ancestry, the generated output was along the lines of...

```
*+.usa-prose>h2,*+.usa-prose>h3 {
    margin-top: 1.7em
}
```

This can be interpreted as "a heading within a prose block, where the prose block itself is preceded by any other content".

This is expected to have a visual impact on headings in prose content. Example, from the ["Typography" page](https://design.login.gov/typography/):

Before|After
---|---
before|after

Given that this styling has existed for quite a while, it may be worth reevaluating if the margins are still desirable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Fix an issue where builds may produce styles in an unpredictable order.
 - Fix an issue where `.usa-display` heading font size was rendered larger than intended.
 - Fix an issue where `dropdownButton` was using invalid CommonJS syntax when imported from ES module entrypoint.
+- Fix an issue where `.usa-prose` heading margins were not applied consistently.
 
 ### Internal
 
--- a/src/scss/components/_typography.scss
+++ b/src/scss/components/_typography.scss
@@ -13,32 +13,37 @@
     letter-spacing: -0.2px;
   }
 
-  h2, h3 {
-    * + & {
-      margin-top: 1.7em;
-    }
+  * + h2,
+  * + h3 {
+    margin-top: 1.7em;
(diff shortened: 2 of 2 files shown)


zachmargolis (approved):
LGTM I like the snug headers

zachmargolis:
Oh wait I got before and after mixed up. I'm a fan of the snug headers, but if this is more consistent with something else, I'm OK with that too

aduth:
> Oh wait I got before and after mixed up. I'm a fan of the snug headers, but if this is more consistent with something else, I'm OK with that too

It's more that the previous styles weren't really doing what we probably expected them to. I don't have a strong preference to one or the other, but it felt like the intention was to have the margins. And if we don't want them, we'd at least want to remove the `* + &` styling.

For context, I encountered it with making changes to the branch at #181 to adapt additional content into the "Images" page, where the heading margins didn't feel quite right in a real scenario:

Current Styles|Margins Corrected
---|---
current|margins corrected

aduth:
I'd also say it's generally more consistent. You can see the headers look decent on the Brand page:

https://design.login.gov/brand/

They look correct there because the sidebar is technically an element preceding `.usa-prose`. But definitely not what I'd expect to be a prerequisite for header margins to be applied 😬 

The Typography page is a bit of an extreme example, where in the real-world there'd usually be paragraph content between headings.

PR #190, opened 2021-02-25
LG-4226: Fix double focus ring

aduth:
This pull request seeks to resolve an issue where focused buttons appear with two focus rings. The button styles included in the design system intend to unset the inherited focus styles from USWDS, but are not of high enough specificity to override the styles.

Reference: https://github.com/uswds/uswds/blob/1f9bb1cc/src/stylesheets/global/_focus.scss#L1-L11

Before|After
---|---
before|after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Bug Fixes
+
+- Fix an issue where focused buttons appear with a double focus ring style.
+
 ## 4.0.0
 
 ### Breaking Changes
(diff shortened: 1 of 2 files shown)


anniehirshman-gsa (commented):
Before/After LGTM! 👍  This applies to all the other types of buttons too?

aduth:
> This applies to all the other types of buttons too?

The fix does apply, yeah. There's a couple custom revisions we'll still need to make for focus styles for the unstyled buttons, since that variant isn't really styled by us at all and appears pretty broken in its current state. I'm working on that right now in a separate branch which will "introduce" unstyled as a new option.

anniehirshman-gsa (approved):
Great thanks!

PR #191, opened 2021-02-25
LG-4226: Improve support for unstyled button

aduth:
The unstyled button variant had previously existed in that it was inherited from USWDS, though it was not documented, and did not always display correctly, particularly in hover, active, and focus states. These changes improve this support, and adds documentation for the unstyled option.

**Screenshot:**

unstyled button

**Live Preview:** https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-lg-4226-document-variants/components/buttons/

### DIFF
--- a/src/scss/components/_buttons.scss
+++ b/src/scss/components/_buttons.scss
@@ -27,6 +27,50 @@
   }
 }
 
+.usa-button--unstyled {
+  position: relative;
+  text-underline-offset: 3px;
+
+  // In USWDS, unstyled button styles are applied last to take precedence over base button styles.
+  // Since we override base button styles, we must re-apply the unstyled button styles.
+  &.usa-button {
+    @include button-unstyled;
+  }
+
+  // Hover colors are already applied by USWDS via `typeset-link`, but this does not include the
+  // button modifier classes.
+  &:hover,
+  &.usa-button--hover {
+    color: color($theme-link-hover-color);
+  }
+
+  &:active,
+  &.usa-button--active {
+    color: color($theme-link-active-color);
+  }
+
+  &:disabled {
+    color: color('disabled');
+  }
+
+  &.usa-button:not([disabled]):focus,
+  &.usa-button:not([disabled]).usa-focus {
+    box-shadow: none;
+
+    &::before {
+      border-radius: .125rem;
+      bottom: -5px;
+      box-shadow: 0 0 0 $theme-focus-width color($theme-focus-color);
+      content: '';
+      left: -5px;
+      pointer-events: none;
+      position: absolute;
+      right: -5px;
+      top: -5px;
+    }
+  }
+}
+
 .usa-button:not([disabled]):focus,
 .usa-button:not([disabled]).usa-focus {
   @include disable-default-focus-styles;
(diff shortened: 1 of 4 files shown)


src/scss/components/_buttons.scss
```diff
@@ -27,6 +27,49 @@
...
+  // button modifier classes.
+  &:hover,
+  &.usa-button--hover {
+    color: color($theme-link-hover-color);
+  }
+
+  &:active,
+  &.usa-button--active {
+    color: color($theme-link-active-color);
+  }
```
aduth, line 50: @anniehirshman-gsa I noticed in the Figma that the hover and active states should be different colors. I think what we have here is the correct _implementation_, as far as using `$theme-link-hover-color` and `$theme-link-active-color`, but it appears they're currently configured to the same color.

I think it might make sense to update that globally so that the link hover and active colors are distinct, if that's the intention. What do you think?

src/scss/components/_buttons.scss
```diff
@@ -27,6 +27,50 @@
...
+
+  &:disabled {
+    color: color('disabled');
+  }
+
+  &.usa-button:not([disabled]):focus,
+  &.usa-button:not([disabled]).usa-focus {
```
zachmargolis, line 57: Just to clarify, this is different from the stanza as below because the `&` means these styles *when they apply to `.usa-button--unstyled`* right, and the one below is when they're on any button?

zachmargolis (approved):
LGTM

src/scss/components/_buttons.scss
aduth: > Just to clarify, this is different from the stanza as below because the `&` means these styles _when they apply to `.usa-button--unstyled`_ right, and the one below is when they're on any button?

Yeah, when compiled it's the difference of:

```css
.usa-button--unstyled.usa-button:not([disabled]):focus {
  /* ... these styles */
}
```

vs.

```css
.usa-button:not([disabled]).usa-focus {
  /* ... default focus styles */
}
```

Where the intention with these styles is to unset and replace the default focus styles, only for the unstyled buttons. _Ideally_ we could just use the default styles, but it was surprisingly difficult to reconcile the focus style box shadow + border radius, hence recreating it with a positioned pseudo-element.

src/scss/components/_buttons.scss
anniehirshman-gsa: Yes I would like to update `$theme-link-hover-color: 'primary-dark';` and `$theme-link-active-color: 'primary-darker';` globally, to match the unstyled button mockups in Figma.  Currently links have a hover color of `primary-darker` and no active color. @nickttng does that sound good to you too?

nickttng: @anniehirshman-gsa Yep, that sounds good

anniehirshman-gsa (approved):
LGTM (with change to hover and active colors in #196 )

aduth:
>  (with change to hover and active colors in #196 )

It might take a few minutes to finish building, but I rebased this branch after merging #196, so the preview should reflect the new hover colors in a few minutes:

https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-lg-4226-document-variants/components/buttons/

PR #192, opened 2021-02-25
LG-4226: Add wide button variants

aduth:
Adds two new button variants:

1. Minimum width occupies full width on mobile, and a minimum width on wider viewports
2. Full width always occupies full width

**Questions:** (cc @anniehirshman-gsa)

1. The Figma includes text "All buttons default to full width for mobile viewports". Should this behavior apply to the unmodified `usa-button` ("flexible") button, or only minimum-width and full-size buttons?
2. The Figma mentions using `btn-wide` as the class for a minimum width button. As implemented here, I chose to follow the BEM convention as `usa-button--wide`. Is that okay?
3. What specific breakpoint should mobile behavior stop taking effect ([see USWDS "Utility breakpoints"](https://designsystem.digital.gov/documentation/settings/#utilities-settings))? As implemented here is `tablet`.
4. I included an explicit `usa-button--full-width` modifier to handle cases like in the IDP where we want full-width buttons on desktop viewports. Is that reasonable to include here, or should that be considered an exceptional case and custom to the IDP?

**Screenshot:**

button widths

**Live Preview:** https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-lg-4226-wide-button/components/buttons/#button-widths

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
   - Before: Hover and active colors are both `primary-darker`.
   - After: Hover is `primary-dark`, and active is `primary-darker`.
 - Improved support for "Unstyled" button variant ([see documentation](https://design.login.gov/components/buttons/))
+- Add two new wide-width button variants `usa-button--wide` and `usa-button--full-width`.
 
 ### Bug Fixes
 
--- a/docs/_components/buttons.md
+++ b/docs/_components/buttons.md
@@ -81,7 +81,7 @@ Default button width for desktop viewports.
 
 ### Minimum width
 
-Use to set a minimum button width for desktop viewports.
+Use `usa-button--wide` to set a minimum button width for desktop viewports.
 
 ```html
 <button class="usa-button usa-button--wide">
@@ -94,6 +94,8 @@ Use to set a minimum button width for desktop viewports.
 ### Full width
 
 All buttons default to full width for mobile viewports.
+
+Use `usa-button--full-width` to set full-width buttons for desktop viewports.
 
 ```html
 <button class="usa-button use-button--full-width">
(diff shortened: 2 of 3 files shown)


zachmargolis (approved):
LGTM

anniehirshman-gsa (approved):
This looks great!  To answer your qs: 

1. The Figma includes text "All buttons default to full width for mobile viewports". Should this behavior apply to the unmodified usa-button ("flexible") button, or only minimum-width and full-size buttons?  **The unmodified "flexible" button should be full width on mobile as well, in alignment with USWDS defaults**
2. The Figma mentions using btn-wide as the class for a minimum width button. As implemented here, I chose to follow the BEM convention as usa-button--wide. Is that okay?  **Yes, whatever class name makes the most sense with our conventions. Updated in Figma**
3. What specific breakpoint should mobile behavior stop taking effect (see USWDS "Utility breakpoints")? As implemented here is tablet. **Yes, tablet (640px) seems to be our current main breakpoint in the IdP between mobile and desktop layouts, but lmk if you catch inconsistencies on this because that would be good to know.** 
4. I included an explicit usa-button--full-width modifier to handle cases like in the IDP where we want full-width buttons on desktop viewports. Is that reasonable to include here, or should that be considered an exceptional case and custom to the IDP?  **Yes this is great! Updated in Figma**

aduth:
> Yes, tablet (640px) seems to be our current main breakpoint in the IdP between mobile and desktop layouts, but lmk if you catch inconsistencies on this because that would be good to know.

I checked on this more closely, and it appears that USWDS actually cuts over at `mobile-lg` for buttons in particular:

https://github.com/uswds/uswds/blob/1f9bb1c/src/stylesheets/elements/_buttons.scss#L24-L26

I assume we'd probably want to match that.

PR #200, opened 2021-03-10
LG-4226: Fix double focus ring on unstyled buttons

aduth:
Regression of #193 (69a13a0342ae871cac8383e138d4b603b45f4d04).

**Why**: For consistency with other buttons and to align to expected design specs, there should only be a single outline on any focussed button.

Before|After
---|---
before|after

**Live Preview:** https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-fix-unstyled-focus/components/buttons/

**Implementation Notes:**

Refactoring in #193 rearranged styles such that [our re-application of the unstyled mixin](https://github.com/18F/identity-style-guide/blob/8a6b48308f27cf7f6b75df54cdc50981f3708722/src/scss/components/_buttons.scss#L15-L19) would reintroduce focus styles through [the mixin's use of link styles](https://github.com/uswds/uswds/blob/0c95fd94ecdb646f360592aa01c55dc386ce407b/src/stylesheets/core/mixins/_button-unstyled.scss#L3). Ultimately, including this mixin should not be necessary, because USWDS handles hover and active unstyled states, but does not currently do so for the modifier classes (see https://github.com/uswds/uswds/pull/4077). Previously, we worked around this by including the unstyled button mixin, which would have a higher specificity than the default button's hover styles. Unfortunately, this also comes with some unexpected consequences, such as the [re-application of the USWDS focus ring](https://github.com/uswds/uswds/blob/0c95fd94ecdb646f360592aa01c55dc386ce407b/src/stylesheets/core/mixins/_typography.scss#L75-L77), which we expect to override in [our own default button styles](https://github.com/18F/identity-style-guide/blob/8a6b48308f27cf7f6b75df54cdc50981f3708722/src/scss/components/_buttons.scss#L4-L8).

This branch includes a bump to the patch version of the package, with the expectation that a new package will be published immediately following merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 5.0.1
+
+### Bug Fixes
+
+- Fix an issue where a second focus ring effect was still applied to the unstyled button variant.
+
 ## 5.0.0
 
 ### Breaking Changes
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "identity-style-guide",
-  "version": "5.0.0",
+  "version": "5.0.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "identity-style-guide",
-  "version": "5.0.0",
+  "version": "5.0.1",
   "description": "The global style of login.gov",
   "main": "./build/cjs/index.js",
   "module": "./build/esm/index.js",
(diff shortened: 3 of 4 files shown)


anniehirshman-gsa (commented):
Focus ring looks good!  Looks like the large unstyled buttons now have left/right padding though, while the regular size unstyled buttons don't. I would vote to remove the padding here and not specify it at the design system level for either size of unstyled buttons (see screenshot in #193 ), but lmk if you feel otherwise

zachmargolis (approved):
LGTM

PR #201, opened 2021-03-22
Buttons: Improve unstyled button styles to visually match link appearance

aduth:
Related Slack conversation: https://gsa-tts.slack.com/archives/CNCGEHG1G/p1616421634001700

**Why**: The use of an unstyled button is intended to take the visual appearance of a link. The proposed changes resolve some discrepancies that currently exist:

- Avoid customizing the underline offset
- Reset font smoothing using `inherit` (see https://github.com/uswds/uswds/pull/4115 for detail)

**Live Preview:** https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-unstyled-buttons-links/components/buttons/

**Screenshot:**

Before|After
---|---
before|after

This was originally observed in the IdP, where an equivalent link and unstyled button did not appear the same. See examples below which show a real link on the left vs. an unstyled button on the right. The IdP customizes the default font smoothing of text, which currently causes a difference in appearance between links and unstyled buttons (see https://github.com/uswds/uswds/pull/4115). Separately, we should consider if this is actually something we should be doing, perhaps as part of the ongoing work to migrate to Public Sans font ([see related guidance](https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/)).

IdP Before|IdP with underline offset fix|IdP with underline offset & smoothing fixes
---|---|---
idp before|idp with underline fix|idp with underline and smoothing fix

### DIFF
--- a/src/scss/components/_buttons.scss
+++ b/src/scss/components/_buttons.scss
@@ -31,7 +31,8 @@
   &.usa-button--active,
   &:disabled,
   &.usa-button--disabled {
-    @include no-knockout-font-smoothing;
+    -moz-osx-font-smoothing: inherit;
+    -webkit-font-smoothing: inherit;
     background-color: transparent;
     box-shadow: none;
     text-decoration: underline;
--- a/src/scss/components/_buttons.scss
+++ b/src/scss/components/_buttons.scss
@@ -19,7 +19,6 @@
   // Since we override base button styles, we must re-apply the unstyled button styles.
   @include button-unstyled;
   position: relative;
-  text-underline-offset: 3px;
 
   // USWDS currently only sets hover and active unstyled button styles for the browser-native
   // states, and not with the modifier classes. It also does not apply these styles to the disabled
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 5.0.2
+
+### Bug Fixes
+
+- Fix an issue where unstyled buttons may appear heavier than an equivalent link.
+
 ## 5.0.1
 
 ### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Bug Fixes
 
-- Fix an issue where unstyled buttons may appear heavier than an equivalent link.
+- Fix the appearance of unstyled buttons to appear visually identical to a link.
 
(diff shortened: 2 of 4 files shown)


aduth:
Interesting that the tests pass 🤔  I would expect visual regressions to be flagged here.

nickttng (approved):

zachmargolis:
> Interesting that the tests pass 🤔 I would expect visual regressions to be flagged here.

I tried running locally, got some timeouts, but also this:

```
 FAIL  test/screenshot.test.js (118.968 s)
  ● screenshot visual regression

    assert.strictEqual(received, expected)

    Expected value to strictly be equal to:
      0
    Received:
      4
    
    Message:
      Expected "/research/1-flow-diagrams/" to visually match the live site.

```

🤷

zachmargolis (approved):
LGTM

aduth:
> but also this:

The diffs should be in your local relative `tmp` folder to take a look at what it might be considering different. `4` seems like a strange result, since I'm pretty sure that number is "total pixel difference", and 4 feels very small. If I had to guess looking at [that particular page](http://localhost:4000/research/1-flow-diagrams/), I'd guess something with how those very-large-images are loaded or scaled.

aduth:
> Interesting that the tests pass 🤔 I would expect visual regressions to be flagged here.

Taking a look locally, I think the issue is that `text-underline-offset` support in Chrome is relatively new ([v87](https://developer.mozilla.org/en-US/docs/Web/CSS/text-underline-offset#browser_compatibility)), and the version of Chromium used in the visual regression tests is a bit older, and therefore the previously styling effectively did nothing because it wasn't supported.

src/scss/components/_buttons.scss
```diff
@@ -31,7 +30,8 @@
   &.usa-button--active,
   &:disabled,
   &.usa-button--disabled {
-    @include no-knockout-font-smoothing;
+    -moz-osx-font-smoothing: inherit;
+    -webkit-font-smoothing: inherit;
```
aduth, line 34: These styles only apply for hover, active, and disabled 🤦  Should have also duplicated above, after the `@include button-unstyled`. Was misled by the fact that this is substituting the behavior of `no-knockout-font-smoothing`.

PR #202, opened 2021-03-23
Apply inherited font smoothing to base unstyled button styles

aduth:
**Why**: These were applied to modifier states for unstyled buttons in #201, but should have also been applied to the base styles, to override USWDS unstyled button styles.


zachmargolis (approved):
LGTM

PR #203, opened 2021-03-23
Use aXe runner for pa11y-ci accessibility test

aduth:
**Why**:

- Consistency with other Login.gov projects ([brochure site](https://github.com/18F/identity-site/blob/544acfd84ce6989710501a409feffa0f358be3ea/package.json#L32), [IdP](https://github.com/18F/identity-idp/blob/ce7608a50ea9f7b8d8c747c63cc0c49549eac4a7/Gemfile#L103), [dashboard](https://github.com/18F/identity-dashboard/blob/33e234da0263723a35eb9cf332f0fb5331bf940c/Gemfile#L75), [PKI](https://github.com/18F/identity-pki/blob/921c14ba9c5c3df83e5768b72f83b1ee9ca8e2df/Gemfile#L45)).
- More capable of finding issues not discovered by HTML_CodeSniffer ([the default runner](https://github.com/pa11y/pa11y#runners)) because it runs in a full browser, as opposed to doing a static analysis of HTML markup.

The second point is further evidenced by the fact that the build is expected to fail with these changes with a number of existing issues on the documentation site.

This work could also facilitate future effort to use `axe-core` directly, as we do in the brochure site, to eliminate the current duplicate dependency on Puppeteer.

### DIFF
--- a/.pa11yci.js
+++ b/.pa11yci.js
@@ -5,5 +5,6 @@ module.exports = {
     },
     "concurrency": 10,
     "standard": "WCAG2AA",
+    "runners": ["axe"],
   }
 }


zachmargolis (approved):
LGTM!

> Aside must not be contained in another landmark

Do we need to fix this error to get this PR to land?

aduth:
> > Aside must not be contained in another landmark
> 
> Do we need to fix this error to get this PR to land?

Yeah, I think I'll plan to leave this one open and create some separate pull requests to fix existing issues. There's quite a few in addition to this one.

aduth:
>Ensure that scrollable region has keyboard access (https://dequeuniversity.com/rules/axe/3.5/scrollable-region-focusable?application=axe)

Our code snippets should be scrollable, since per the resource above, a keyboard-only user can't read the content otherwise ([SC 2.1.1](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)).

It's a bit tricky given how we want to rely on Jekyll + Rouge in many of our projects, which as best I can tell doesn't provide much customization for adding a `tabindex="0"` attribute or wrapping element.

A few thoughts:

- Maybe we'd rather have an explicit component wrapper for code snippets, like `<figure class="lg-code">`. This would be more conventional in the design system vs. global element-level styles or assumptions about particular frameworks like Jekyll. But it's also a breaking change that requires downstream projects to find a way to add the wrapping element, which is not straight-forward if they're relying on Jekyll highlighting.
- Maybe we don't let the content area overflow, and instead let the text wrap. This looks reasonable in some cases, and not so much in others. It may also be simpler to navigate content this way, vs. scrollable regions. This is how [the Deque page referenced above](https://dequeuniversity.com/rules/axe/3.5/scrollable-region-focusable) implements their own code snippets. It's also something we can customize purely within the existing `pre` and/or Jekyll + Rouge `.highlight` CSS styles.
- Maybe there's an option to support both, where `pre.highlight` continues to exist as it does today, but an explicit component element option (`<figure class="lg-code">`) exists as well as an alternative. The problem with this approach is that this issue of focusable scrollable elements would continue to affect any project that relies on that Jekyll + Rouge `pre.highlight` behavior.

aduth:
Rather than let this sit open, I'm going to close the pull request and use the branch as a base against which to propose accessibility improvements, since we'll need to sort out the existing issues first before changing the runner.

PR #204, opened 2021-03-25
Add custom components to LGDS

jmdembe:
Why:
- To document customizations made between LGDS and USWDS

### DIFF
--- a/src/scss/components/_sidenav.scss
+++ b/src/scss/components/_sidenav.scss
@@ -38,3 +38,7 @@
   color: $theme-color-primary-dark;
   font-weight: bold;
 }
+
+.usa-sidenav__sublist .usa-sidenav__item {
+  border-top: 0;
+}
--- a/src/scss/components/_sidenav.scss
+++ b/src/scss/components/_sidenav.scss
@@ -41,6 +41,6 @@
 
 .usa-sidenav .usa-current {
   &:hover {
-      color: #205493;
+      color: color('primary-dark');
   }
 }
--- a/src/scss/components/_sidenav.scss
+++ b/src/scss/components/_sidenav.scss
@@ -37,7 +37,7 @@
 
 li > .active {
   color: $theme-color-primary-dark;
-  font-weight: bold;  
+  font-weight: bold;
 }
 
 .sidenav {
--- a/src/scss/components/_sidenav.scss
+++ b/src/scss/components/_sidenav.scss
@@ -33,9 +33,6 @@
 
 .usa-sidenav__sublist .usa-current {
   @include remove-bar;
-  background-color: color('white');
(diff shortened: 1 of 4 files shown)


src/scss/components/_sidenav.scss
```diff
@@ -34,3 +34,14 @@
   padding-top: units(1.5);
   padding-bottom: units(1.5);
 }
+
+li > .active {
+  color: $theme-color-primary-dark;
+  font-weight: bold;
+}
```
aduth, line 41: Since the Login design system extends the USWDS side navigation component, what do you think about applying the styling to those existing classes, rather than creating new ones?

We already add some custom styles above for `usa-current`, so we might have to "undo" those to get the effect we want with the `.active` class here:

```suggestion
.usa-sidenav__sublist .usa-current {
  @include remove-bar;
  background-color: color('white');
  color: $theme-color-primary-dark;
  font-weight: bold;
}
```

src/scss/components/_sidenav.scss
```diff
@@ -34,3 +34,14 @@
...
+
+.sidenav {
+  ul {
+    border: none;
+  }
+}
```
aduth, line 47: In the `.usa-sidenav` styles above, we add a border style. If the expectation is that the LGDS sidenav has no border, could we remove it there?

jmdembe: Yes. I will remove that border.

aduth:
In the reference Figma, I see no border between the child elements of a side navigation. Should that be reflected here?

Figma|Current Brach
---|---
image|image

src/scss/components/_sidenav.scss
```diff
@@ -34,3 +30,14 @@
   padding-top: units(1.5);
   padding-bottom: units(1.5);
 }
+
+.usa-sidenav__sublist .usa-current {
+  @include remove-bar;
+  background-color: color('white');
+  color: $theme-color-primary-dark;
+  font-weight: bold;
```
aduth, line 38: We might not need these anymore? It looks like with your other recent changes, these styles are already effectively inherited from styles defined above.

```suggestion
```

aduth (approved):
Looks good 👍

src/scss/components/_sidenav.scss
```diff
@@ -34,3 +30,17 @@
...
+  @include remove-bar;
+}
+
+.usa-sidenav__sublist .usa-sidenav__item {
+  border-top: 0;
+}
+
+.usa-sidenav .usa-current {
+  &:hover {
+      color: #205493;
```
aduth, line 44: Can we use the [corresponding color token](https://design.login.gov/color/) instead of hardcoding the hex?

```suggestion
      color: color('primary-dark');
```

PR #205, opened 2021-03-29
Reduce border width on default outline button

aduth:
**Why**: Per Figma design reference, the outline button should only have a border width of 1px, not 2px. The "Big" outline button should still have a 2px border.

**Live Preview:** https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-outline-button-border-width/components/buttons/

**Screenshot:**

_Design System:_

Before|After
---|---
before|after

_IdP:_

Before|After
---|---
before|after

**Implementation Notes:**

USWDS allows customization of the stroke width of an outline button, but does not allow for different values between default and big versions of the button. Implementing this requires duplicating the base styles with a custom "big" stroke width, computed here as double the value of the new default stroke width of `1px`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 5.0.4
+
+### Bug Fixes
+
+- Reduce the border width of the normal outline button to the intended width of `1px` (previously `2px`). The big variant of the outline button is unaffected by this change, and remains `2px`.
+
 ## 5.0.3
 
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


anniehirshman-gsa (approved):
LGTM 👍

zachmargolis (approved):
LGTM

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 5.0.4
```
aduth, line 1: Gonna revert this to "Unreleased" since we can probably ship this together with #204.

```suggestion
## Unreleased
```

PR #206, opened 2021-03-29
Consolidate unstyled button font smoothing styles

aduth:
Related: https://github.com/18F/identity-idp/pull/4825#discussion_r600735686

**Why**:

- Account for missing `usa-button--unstyled:hover` and `usa-button--unstyled:active`
- Single point to remove in the future, as noted in the preceding comment


zachmargolis (approved):
LGTM

aduth:
I think visual regression got confused with #214 since it runs against the live site, and the live site update was probably delayed around the latest rebase here.

PR #211, opened 2021-04-05
Fix build setup to be compatible with latest Federalist release

aduth:
**Why**: Federalist will install all dependencies for us. Previously we needed to re-run the NPM install in order to install development dependencies in addition to production dependencies. As of the Federalist release this past Friday, it will install all dependencies, so this will no longer be necessary.

**Note:** All Federalist builds will fail until this pull request is merged.

Related discussion: https://gsa-tts.slack.com/archives/C1NUUGTT5/p1617287819021500

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,12 +8,6 @@ ifdef SITE_PREFIX
 	OUTPUT_DIR = ./_site
 endif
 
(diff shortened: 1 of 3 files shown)


zachmargolis (approved):
LGTM

PR #212, opened 2021-04-06
Load utility classes after components

aduth:
**Why**: To match USWDS intended load order of stylesheets, to allow utility classes to take precedence over component styles, and to facilitate a potential future use of [USWDS 2.11.0 package subsetting](https://designsystem.digital.gov/about/releases/#v2.10.0%20Beta%200-intermediate-package-subsetting).

USWDS [loads utility stylesheets after components](https://github.com/uswds/uswds/blob/develop/src/stylesheets/uswds.scss), allowing for those utility classes to apply usually as expected as an override to the component style selectors of equal specificity, as an alternative to resorting to [`$utilities-use-important`](https://github.com/uswds/uswds/blob/1908d1391bc59410624ca1934cc70b7404e8f443/src/stylesheets/theme/_uswds-theme-utilities.scss#L20).

Previously, we imported all USWDS styles prior to loading our own component overrides. Thus, the component overrides could take precedence over the utility classes. For example, it was not possible to use [margin utility classes](https://designsystem.digital.gov/utilities/margin-and-padding/) with the [unstyled button](https://design.login.gov/components/buttons/#unstyled) (observed in https://github.com/18F/identity-idp/pull/4869). 

Note that depending on the selector used for a component's styles, a utility class may still "lose" to a component's own styles. If this proves to be problematic (or is argued to be problematic now), it may be worth reevaluating to consider using the `$utilities-use-important` theme setting value, which applies [`!important`](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#the_!important_exception) to all utility classes.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -10,7 +10,6 @@ const gulpif = require('gulp-if');
 const sass = require('gulp-sass');
 const gulpStylelint = require('gulp-stylelint');
 const sourcemaps = require('gulp-sourcemaps');
-const uswds = require('uswds-gulp/config/uswds');
 const browserify = require('browserify');
 const babel = require('gulp-babel');
 const source = require('vinyl-source-stream');
@@ -124,7 +123,7 @@ gulp.task('build-sass', () => {
     .pipe(sourcemaps.init({ largeFile: true }))
     .pipe(
       sass({
-        includePaths: [PROJECT_SASS_SRC, `${uswds}/scss`, `${uswds}/scss/packages`],
+        includePaths: [PROJECT_SASS_SRC, 'node_modules'],
       }).on('error', notificationOptions.handler),
     )
     .pipe(postcss(plugins))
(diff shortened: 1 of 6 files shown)


zachmargolis (approved):
LGTM

gulpfile.js
```diff
@@ -146,18 +145,17 @@ const underscorePrefix = () => gulpif((f) => f.basename[0] !== '_', rename({ pre
...
 );
 
 gulp.task('copy-uswds-scss', () =>
   gulp
-    .src([`${uswds}/scss/**/*.scss`])
+    .src(['node_modules/uswds/dist/scss/**/*.scss'])
     .pipe(replaceUrls())
     .pipe(underscorePrefix())
-    .pipe(gulp.dest(`${SCSS_DEST}/uswds`)),
+    .pipe(gulp.dest(`${SCSS_DEST}/uswds/dist/scss`)),
```
aduth, line 158: This does change the output path for USWDS files from `dist/assets/scss/uwds` to `dist/assets/scss/uswds/dist/scss`. It's not ideal, but avoids some of the "magic" from the renamed import removed above in this file, and it's generally easier to resolve with `includePaths`. I'd hoped this wouldn't be a problem for consumers, but I see now we're referencing some of those paths in the IDP:

https://github.com/18F/identity-idp/blob/c1a0269/config/webpack/environment.js#L39-L41

It wouldn't be much of a problem to update those, and I still don't necessarily consider this a "public" interface, but I'll see if I can adjust this to avoid changing the output directory.

aduth: Symlink seems like a reasonable option: 4c62728

The other obvious option would be to do the same `cp` of USWDS source files in development that occurs at build-time, which is probably a little slower, but simplifies things by ensuring it behaves the same in development and in the built distributable.

PR #214, opened 2021-04-28
Add package entrypoints for granular stylesheet loading

aduth:
**Why**: To limit or control the loading of styles.

Use-cases:

1. Avoid specificity conflicts between utility classes and a consuming project's custom components
2. Load only required mixins and functions to support CSS code splitting

This also mirrors USWDS packages, with two main differences:

- Simplifies naming of `uswds-components` and `uswds-utilities` to just `components` and `utilities`
- Extracts `utilities-required` from `utilities` to allow non-rendering import of utility mixins.

See:

- https://designsystem.digital.gov/components/packages/
- https://github.com/uswds/uswds/blob/develop/src/stylesheets/uswds.scss

Next steps:

Per the above use-cases, this is useful for `identity-idp` and the following diff would help toward resolving existing specificity conflicts and simplify code splitting usage:

```diff
diff --git a/app/assets/stylesheets/application.css.scss b/app/assets/stylesheets/application.css.scss
index 8576adb31..9a5ab78d6 100644
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -1,6 +1,9 @@
 @import 'variables/colors';
 @import 'variables/app';
 @import 'vendor';
-@import 'identity-style-guide/dist/assets/scss/styles';
+@import 'identity-style-guide/dist/assets/scss/packages/required';
+@import 'identity-style-guide/dist/assets/scss/packages/global';
+@import 'identity-style-guide/dist/assets/scss/packages/components';
 @import 'components/all';
 @import 'print';
+@import 'identity-style-guide/dist/assets/scss/packages/utilities';
diff --git a/config/webpack/environment.js b/config/webpack/environment.js
index c55bb013f..65fd2682b 100644
--- a/config/webpack/environment.js
+++ b/config/webpack/environment.js
@@ -25,19 +25,8 @@ const sassLoader = environment.loaders.get('sass');
 sassLoader.use.find(({ loader }) => loader === 'sass-loader').options.additionalData = `
 $font-path: '~identity-style-guide/dist/assets/fonts';
 $image-path: '~identity-style-guide/dist/assets/img';
-@import '~identity-style-guide/dist/assets/scss/functions/asset-path';
-@import '~identity-style-guide/dist/assets/scss/functions/focus';
-@import '~identity-style-guide/dist/assets/scss/uswds-theme/custom-styles';
-@import '~identity-style-guide/dist/assets/scss/uswds-theme/general';
-@import '~identity-style-guide/dist/assets/scss/uswds-theme/typography';
-@import '~identity-style-guide/dist/assets/scss/uswds-theme/spacing';
-@import '~identity-style-guide/dist/assets/scss/uswds-theme/color';
-@import '~identity-style-guide/dist/assets/scss/uswds-theme/utilities';
-@import '~identity-style-guide/dist/assets/scss/uswds-theme/components';
-@import '~identity-style-guide/dist/assets/scss/uswds/packages/required';
-@import '~identity-style-guide/dist/assets/scss/uswds/utilities/palettes/all';
-@import '~identity-style-guide/dist/assets/scss/uswds/utilities/rules/all';
-@import '~identity-style-guide/dist/assets/scss/uswds/utilities/rules/package';`;
+@import '~identity-style-guide/dist/assets/scss/packages/required';
+@import '~identity-style-guide/dist/assets/scss/packages/utilities-required';`;
 
 sassLoader.use.find(({ loader }) => loader === 'css-loader').options.sourceMap = false;
```

Example conflicts:

- https://github.com/18F/identity-idp/blob/0a81dd7/app/assets/stylesheets/components/_btn.scss#L38
- https://github.com/18F/identity-idp/blob/0a81dd7/app/views/idv/shared/_error.html.erb#L40
   - Should use `margin-top-5` but currently conflicts with custom component's [own margins](https://github.com/18F/identity-idp/blob/0a81dd7441038ae7992d6923b7b73138ab91272f/app/assets/stylesheets/components/_troubleshooting-options.scss#L2)

### DIFF
--- a/src/scss/packages/_utilities-required.scss
+++ b/src/scss/packages/_utilities-required.scss
@@ -1,0 +1,3 @@
+@import '../uswds/utilities/utility-fonts';
+@import '../uswds/utilities/palettes/all';
+@import '../uswds/utilities/rules/all';
--- a/src/scss/packages/_utilities-required.scss
+++ b/src/scss/packages/_utilities-required.scss
@@ -1,3 +1,0 @@
-@import '../uswds/utilities/utility-fonts';
-@import '../uswds/utilities/palettes/all';
-@import '../uswds/utilities/rules/all';
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### New Features
+
+- Add support for SASS package entrypoints to limit or control loading of styles, [similar to USWDS components](https://designsystem.digital.gov/components/packages/). Available packages are `packages/required`, `packages/global`, `packages/components`, `packages/utilities-required`, and `packages/utilities` (includes `packages/utilities-required`).
+
 ### Bug Fixes
 
 - Reduce the border width of the normal outline button to the intended width of `1px` (previously `2px`). The big variant of the outline button is unaffected by this change, and remains `2px`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### New Features
 
-- Add support for SASS package entrypoints to limit or control loading of styles, [similar to USWDS components](https://designsystem.digital.gov/components/packages/). Available packages are `packages/required`, `packages/global`, `packages/components`, `packages/utilities-required`, and `packages/utilities` (includes `packages/utilities-required`).
+- Add support for SASS package entrypoints to limit or control loading of styles, [similar to USWDS components](https://designsystem.digital.gov/components/packages/). Available packages are `packages/required`, `packages/global`, `packages/components`, and `packages/utilities`.
 
 ### Bug Fixes
 
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -75,6 +75,17 @@ Use the original source scss files if your project will be extending styles â€”Â
 @import 'node_modules/identity-style-guide/dist/assets/scss/styles';
 ```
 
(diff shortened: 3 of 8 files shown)


src/scss/packages/_utilities-required.scss
```diff
@@ -0,0 +1,3 @@
+@import '../uswds/utilities/utility-fonts';
```
aduth, line 1: I could also see an argument for just putting this all into `_required.scss`.

zachmargolis: since this file has required in its name...seems like a good fit to me

zachmargolis (approved):
LGTM

src/scss/packages/_utilities-required.scss
aduth: I guess a main counter-point would be that utilities are a prime candidate for something that someone would opt not to include if they're cherry-picking packages. But on the other hand, this file doesn't actually "render" anything on its own in the sense that it would add styles to an otherwise-small CSS bundle... it just makes those utility mixins/functions available.

I think I'll move to `_required.scss`, since it makes things a fair bit simpler, and we also don't have to worry about how to document this or describe it in the changelog.

aduth: Updated in 12ac3da

aduth:
Test failure is visual regression for expected content changes to [Usage page](https://design.login.gov/usage/).

PR #219, opened 2021-05-18
Upgrade USWDS to latest (from 2.9.0 to 2.11.2)

aduth:
**Why**: Incorporate upstream bug fixes, and get access to: new icon set (2.10.0), tile radio inputs (2.10.0).

Full release notes: https://designsystem.digital.gov/about/releases/

Live preview: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-uswds-2-11/

**Implementation Notes:**

Changes largely include:

- Fixing components to accommodate upstream changes: Accordion, alert, sidenav
- Removing bug fixes which have been incorporated upstream
- Adding newly available components
- Reconciling tile checkbox and radio inputs with our custom set
   - Removing "large" in favor of being the "tile" customization
   - Updating "bordered" class naming for consistency
   - Generally improving styling per Figma designs


src/scss/components/_banner.scss
```diff
@@ -42,7 +47,7 @@
   }
 
   &,
-  &[aria-expanded=true] {
+  &[aria-expanded='true'] {
```
zachmargolis, line 50: this is a style thing right? there aren't "types" in CSS selectors where this would matter?

zachmargolis (approved):
LGTM

src/scss/components/_banner.scss
aduth: > this is a style thing right? there aren't "types" in CSS selectors where this would matter?

Right, just styling. Prompted for consistency's sake due to these intending to explicitly override selector styles from USWDS, which include quotes (though USWDS uses double quotes).

https://github.com/uswds/uswds/blob/2650f15/src/stylesheets/components/_banner.scss#L251

aduth:
@anniehirshman-gsa @nickttng No rush, but would one of you mind taking a look at these changes from a UX perspective? Of particular note are the changes to the checkbox and radio styles, which incorporate new "bordered" and "tile" appearances from USWDS 2.10.0 and from updated Figma designs.

Live preview:
- Before: https://design.login.gov/components/form-fields/#radio-buttons
- After: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-uswds-2-11/components/form-fields/#radio-buttons

Screenshots:

Component|Before|After
---|---|---
Radio|b-radio-before|b-radio-after
Checkbox|b-checkbox-before|b-checkbox-after
Bordered|b-radio-bordered-before|b-radio-bordered-after
Tile (previously "Large")|b-radio-tile-before|b-radio-tile-after



And a few quirks to note:

1. The Figma designs include a half-pixel gap between the focused checkbox borders, which looks decent enough in high pixel-density displays, but a bit blurred in `@1x` displays. You may not see much of a difference in the preview below if you don't have a high pixel-density display.

1x|2x
---|---
b-checkbox-after|b-checkbox-after-2x

2. As designed, the border for the unselected, focused radio button would not have a gap between the radio itself and the focus border, at least in using the base styles from "unselected" and "selected, focus" as reference. The way I chose to address this was to keep the inset border that had existed previously, where the difference is the final size of the unselected radio button is 14px instead of 15px.

As designed|With inset border
---|---
b-radio-after-outset|b-radio-after

nickttng:
@aduth Generally looks good. About the new focus state border, I'm thinking that it might be too subtle visually. It could almost be overlooked.

@anniehirshman-gsa What do you think? Would it benefit from having a slightly thicker border or something more noticeable?

aduth:
> @aduth Generally looks good. About the new focus state border, I'm thinking that it might be too subtle visually. It could almost be overlooked.
> 
> @anniehirshman-gsa What do you think? Would it benefit from having a slightly thicker border or something more noticeable?

For context, this is also something I had to explicitly override as part of the work here for checkboxes and radio buttons specifically, to reduce from the default 2px border down to 1px.

aduth:
> Currently, our other form inputs (buttons, text fields) have a 1px blue focus ring, so I'm fine with that here too for now. But I'd love to see us beef up the focus ring more to match USWDS... but I would vote to do it across all our interactive components (actually created a ticket awhile back, LG-4238, to make that improvement).

Our standard focus ring is `2px`, compared to [USWDS' `4px` (`0.25rem`)](https://github.com/uswds/uswds/blob/2650f15/src/stylesheets/theme/_uswds-theme-general.scss#L109).

https://github.com/18F/identity-style-guide/blob/879c339c28a30634aae984791e839ab199171d12/src/scss/uswds-theme/_general.scss#L81

In which case, if it's consistency we're after, and preferring a beefier width, I could implement the radio / checkbox with the 2px focus?

anniehirshman-gsa:
Whoops you are totally right of course! 

So yes, `2 px` for the focus ring for now sounds good (and then we can change the default style to `4 px`/`.25 rem` another day

aduth:
Pushed changes in d666c54 to use the default (1px offset, 2px width) focus ring for radios.

Before|After
---|---
Screen Shot 2021-05-24 at 4 18 50 PM|Screen Shot 2021-05-24 at 4 19 00 PM

anniehirshman (approved):
LGTM!

PR #220, opened 2021-05-18
Update .nvmrc to use latest LTS release

aduth:
**Why**:

- Fix Federalist builds, which no longer support Node 10
- Avoid pinning to a specific version to reduce future maintenance overhead
- Use a version of Node which hasn't fallen outside the maintenance calendar, to avoid missing critical bug fixes or encountering other conflicts

See: https://nodejs.org/en/about/releases/

We may also want to consider bumping `engines` in `package.json`:

https://github.com/18F/identity-style-guide/blob/9d919460b4c3cf03c24e5b7573348df7d3fdff1a/package.json#L53-L56

...but this means a different thing, as it prescribes the version of Node expected to be present in consuming projects. And since the IDP [indirectly supports `>=12`](https://github.com/18F/identity-idp/blob/57f8bd8f6fefa44798fe17349c8efe613072f9f1/package.json#L5-L8), it would be a mismatch to set this higher at `>=14`. And, strictly speaking, since the code is transpiled, it may be reasonable to expect that it runs in older versions.


zachmargolis (approved):
LGTM

PR #225, opened 2021-06-16
Fix mobile primary navigation styles

aduth:
**Why**: The USWDS upgrade in #219 changed default padding for most navigation menus. This was corrected in the side navigation, but mobile primary navigation still appeared more condensed than previously. Additionally, default scaling for the navigation close button was changed in USWDS to be much larger due to a change to use the new icon set, causing the icon to appear very large. The design system was not using our custom version of the icon anyways. This has been corrected.

Before|After
---|---
image|image

Open questions:

- _Is this what we want the menu to look like?_ I couldn't find a canonical reference for what the mobile menu should look like. The adjustments here are a best compromise of restoring previous expectations. This could be a good opportunity to make updates if we wanted.
- _Do we want to use the USWDS icon?_ Expanding on the above note about USWDS transitioning to new icon set for the mobile "Close" button, it could help simplify our overrides if we use the USWDS icon. That said, while it is very similar in appearance to our custom icon, it's not exactly the same; specifically, there's a slight rounding to the linecaps in our version.
   - [USWDS "Close"](https://raw.githubusercontent.com/uswds/uswds/develop/src/img/usa-icons/close.svg)
   - [Login.gov "Close"](https://raw.githubusercontent.com/18F/identity-style-guide/9ebc78f52ea932c315d3102d6bbbbb0edefb4052/src/img/close.svg)

### DIFF
--- a/src/scss/components/_navbar.scss
+++ b/src/scss/components/_navbar.scss
@@ -95,6 +95,13 @@ $header-height: 10;
         color: color('ink');
       }
     }
+
+    .usa-nav__primary-item {
+      > a,
+      > .usa-nav__link {
+        @include u-padding-y(2);
+      }
+    }
   }
 
   @include at-media($theme-header-min-width) {
@@ -120,10 +127,6 @@ $header-height: 10;
       .usa-current {
         @include add-bar(0.5, 'secondary', 'bottom', 0, 1.5, 0);
         font-weight: font-weight('bold');
-
-        &:hover {
-          color: color('primary-dark');
-        }
       }
     }
 
@@ -281,12 +284,15 @@ $header-height: 10;
   }
 }
 
-// Navigation menu
+// Navigation close button
 // ---------------------------------
 
-.usa-menu__btn {
-  height: units($header-height);
+.usa-nav__close img {
+  width: units(1.5);
 }
+
+// Navigation menu
+// ---------------------------------
 
 .usa-logo__img {
   @include at-media-max($theme-header-min-width) {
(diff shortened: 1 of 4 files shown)


src/scss/components/_navbar.scss
```diff
@@ -120,10 +127,6 @@ $header-height: 10;
       .usa-current {
         @include add-bar(0.5, 'secondary', 'bottom', 0, 1.5, 0);
         font-weight: font-weight('bold');
-
-        &:hover {
-          color: color('primary-dark');
-        }
```
aduth, line 126: Not strictly related to the changes, but these styles are redundant with those above:

https://github.com/18F/identity-style-guide/blob/main/src/scss/components/_navbar.scss#L109-L112

src/scss/components/_sidenav.scss
```diff
@@ -27,8 +27,6 @@
 
 .usa-sidenav__sublist a:not(.usa-button) {
   padding-left: units(4);
-  padding-top: units(1.5);
-  padding-bottom: units(1.5);
```
aduth, line 31: Again not relevant for primary navigation, but in trying to find a canonical reference for the menu, I noticed a small difference in our sublist styling from the Figma reference. Previously, sublist items had smaller padding, but in the [Figma design](https://www.figma.com/file/kXLUewllU5YoQ6xG2fhwOM/Login-Design-System-LGDS-Figma-Library?node-id=4%3A2089) they're identical padding.

<table><thead><tr><th>Before</th><th>After</th><tr></thead><tbody><tr><td valign="top">before</td><td valign="top">after</td></tr></tbody></table>

zachmargolis (approved):
LGTM

aduth:
Visual regression appears to be attributable to the changes described at https://github.com/18F/identity-style-guide/pull/225#discussion_r652702027.

PR #229, opened 2021-07-14
Remove custom Accordion close button element

aduth:
**Why**:

Maintenance:
- Limit divergence from USWDS, facilitating ongoing maintenance and reducing potential confusion introduced by enhancements exclusive to the login.gov Design System.

Accessibility:
- The button's behavior is duplicative with the toggle, where the additional tab stop may be excessive as repeated content.
- Collapsing an accordion by the close button causes a focus loss ([SC 2.4.7](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-visible.html), [SC 2.4.3](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html))
- The relationship between the button and controlled content is not clearly indicated (e.g. [`aria-controls`](https://www.w3.org/TR/wai-aria-1.1/#aria-controls))
- Current button examples inherit semantics of a form submission button by absence of explicit [`type` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type), despite the fact that the button is not intended for form submission.


zachmargolis (approved):
LGTM

PR #233, opened 2021-07-21
Adding Security Experience Principles to the design system

juliaelman:
This pull request does a few things:

1. Creates a `Utilities` section of the design system to capture things like Colors, Imagery, Typography to be used for Login.gov
2. Adds our Security Experience Principles that outlines how the User Experience Team on Login.gov works

PR #234, opened 2021-07-22
Fix integration tests for Color page

aduth:
**Why**: The URL changed from `/color/` to `/utilities/color/` in #233. This updates the integration test accordingly, so that it doesn't fail.


zachmargolis (approved):
LGTM

PR #236, opened 2021-08-09
Refactor SASS variables to only include overrides

aduth:
**Why**: To reduce the potential for drift between the project and the upstream USWDS, to make it easier to identify our customizations, to eliminate inaccuracies and unused variables, and to make the files more easily grokkable via their smaller size.

The intention is that **there should be no perceivable, user-facing impact** of these changes.

An alternative to this approach is to list all variables, regardless of whether we change them from the default. This has some advantages in that it is easier to understand which variables exist to be customized, and to avoid unintended changes if USWDS were to change any defaults in the future. Conversely, this is much harder to maintain, harder to identify which customizations we've made, and the potential for new inherited defaults from upstream releases isn't necessarily a _bad thing_.

It may be easiest to review with whitespace changes hidden: https://github.com/18F/identity-style-guide/pull/236/files?w=1

As reference, the approach for these changes was to compare these files against [the default USWDS settings](https://github.com/uswds/uswds/tree/develop/src/stylesheets/settings), updating any inaccuracies and removing variable assignments where our variables matched the defaults from USWDS.


src/scss/components/_verification-badge.scss
```diff
@@ -2,7 +2,7 @@
 
   background-color: map-get($site-palette, 'site-white');
   border: units(1px) solid map-get($site-palette, 'site-green');
-  border-radius: $theme-border-radius-xxl;
```
aduth, line 5: "XXL" is not a size for border radius in USWDS. This usage must have been intended to consider it as a custom addition. But it's not as simple as adding just the variable, since the size tokens are expected to be usable from within e.g. [utility mixins](https://designsystem.digital.gov/utilities/border/#utility-radius). Code like `@include u-radius('xxl');` doesn't work otherwise. I'm not sure if it's intended to be possible to override these from the [default radius settings](https://github.com/uswds/uswds/blob/0b469e41fa9f6a947ffcd546b79f3a8aec0002ce/src/stylesheets/core/_variables.scss#L598-L617). In any case, since it wasn't really working correctly aside from direct variable reference, and since we don't reference the variable in any of our other projects, I figured the easiest decision is to eliminate the variable and inline its value.

zachmargolis: +1 💊

zachmargolis (approved):
> It may be easiest to review with whitespace changes hidden: https://github.com/18F/identity-style-guide/pull/236/files?w=1

Hiding whitespace changes made it much easier, this is deceptively simple now

Thanks for this, LGTM!

src/scss/components/_verification-badge.scss
aduth: > In practice, this is likely a candidate for the ["pill" style radius](https://designsystem.digital.gov/utilities/border/#utility-radius) anyways 🤔 I may just update it to use that.

Updated in f8e1e63.

PR #237, opened 2021-08-10
Refactor utilities settings to include only overrides

aduth:
Related: #236

This is the same as #236. It was missed as part of that work and should have been included there. That said, it may be better this way as a standalone unit of change, since it's a very large file.

Likely easiest to review the _after_ state of the file, as there's very little which remains as being overridden from the USWDS defaults: https://github.com/18F/identity-style-guide/blob/aduth-prune-util-vars/src/scss/uswds-theme/_utilities.scss

For comparison: https://github.com/uswds/uswds/blob/develop/src/stylesheets/settings/_settings-utilities.scss


zachmargolis (approved):
so much red! LGTM

PR #239, opened 2021-08-17
Checkbox / Radio Button: Improve HTML whitespace sensitivity

aduth:
Related: https://github.com/uswds/uswds/pull/4286
Context: https://github.com/18F/identity-idp/pull/5279#discussion_r686146591

This pull request seeks to maintain the existing visual appearance of the Checkbox and Radio Button components, while improving developer ergonomics regarding HTML whitespace sensitivity.

See additional details at https://github.com/uswds/uswds/pull/4286 .

The expectation is that there should be no user-facing visual effect of these changes, hopefully demonstrated by a passing visual regression test.

This does not incorporate [other design revisions proposed for the checkbox and radio component](https://gsa-tts.slack.com/archives/C01JVKYE4KD/p1628869289001500), which will be implemented separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Improvements
+
+- Whitespace in checkbox and radio label markup will no longer cause the label text to be misplaced.
+
 ## 6.0.0
 
 ### Breaking Changes
--- a/docs/_components/form-fields.md
+++ b/docs/_components/form-fields.md
@@ -146,23 +146,38 @@ Three styles of radio buttons are provided.
   <ul class="usa-input-list">
     <li>
       <input id="a8c1" class="usa-radio__input usa-radio__input--tile" type="radio" name="a8c1">
-      <label for="a8c1" class="usa-radio__label">Option - Unselected<span class="usa-radio__label-description">Support text</span></label>
+      <label for="a8c1" class="usa-radio__label">
+        Option - Unselected
+        <span class="usa-radio__label-description">Support text</span>
+      </label>
     </li>
     <li>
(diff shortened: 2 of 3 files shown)


aduth:
> The expectation is that there should be no user-facing visual effect of these changes, hopefully demonstrated by a passing visual regression test.

I lied... the horizontal offset of the checkboxes changes ever-so-slightly. Checking out the [artifacts for the build](https://app.circleci.com/pipelines/github/18F/identity-style-guide/725/workflows/c6332324-ca6b-4583-be0d-c2332bfee5e1/jobs/1681/artifacts) (specifically `-diff.png`) gives a good overview of the changes. I might be able to get this back to its original styling, but in retrospect I feel this is a bit more faithful to the original Figma designs, as far as the checkboxes and radios appearing at the leftmost edge, with a consistent right margin of 0.5rem. It gives a slightly imbalanced appearance in the bordered and tile variations, but this exists in the Figma design as well. Curious @nickttng @anniehirshman-gsa @cdelivuk if you'd have thoughts here.

Live demo: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-tile-checkbox-label-size/components/form-fields/#radio-buttons

zachmargolis (approved):
LGTM, thanks for highlighting the visual diff. If it's OK with UX team it's OK with me!

anniehirshman-gsa (approved):
This change looks good to me as well... I hadn't noticed that the regular checkboxes and radio buttons were offset by `4px`, but agree that flush left makes more sense. Thanks for catching!

aduth:
Noting that the upstream https://github.com/uswds/uswds/pull/4286 is now merged, we'll need to come back and revisit this in the future to reconcile redundancies.

PR #242, opened 2021-08-25
Increase padding and font size for tile checkbox

aduth:
Related: https://github.com/18F/identity-idp/pull/5279#pullrequestreview-727676760
Slack discussion: https://gsa-tts.slack.com/archives/C01JVKYE4KD/p1628869289001500

Implements padding and font size for tile variants of checkbox and radio buttons to align to Figma designs.

**Screenshots:**

Before|After
---|---
image|image

**Live Preview:**

- Before: https://design.login.gov/components/form-fields/#tile
- After: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-check-tile-padding/components/form-fields/#tile

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Whitespace in checkbox and radio label markup will no longer cause the label text to be misplaced.
 - Margins surrounding an external link icon have been increased slightly.
+- Padding and label font size for tile checkboxes and radio buttons have increased slightly.
 
 ## 6.0.0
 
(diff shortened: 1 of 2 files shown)


zachmargolis (approved):
Visually, the padding looks a little too big to me? But if it matches the designs that's good enough for me 👍

aduth:
> Visually, the padding looks a little too big to me? But if it matches the designs that's good enough for me 👍

Yeah, there was [some prior discussion about this on Slack](https://gsa-tts.slack.com/archives/C01JVKYE4KD/p1628869569002000?thread_ts=1628869289.001500&cid=C01JVKYE4KD). I'll defer to @anniehirshman-gsa or @cdelivuk on the final call, since maybe it makes a difference to see it "live". One thing I'll note is that it might appear a bit excessive in the examples only because there's not much supporting text, whereas in the real world we'll typically have a longer description.

For example:

image

PR #245, opened 2021-09-03
Use login.gov shared ESLint configuration

aduth:
Related: https://github.com/18F/identity-idp/pull/5359, https://github.com/18F/identity-reporting/pull/12

**Why:** So that we can use a common standard ruleset across all repositories where JavaScript is used.

Observations:

- This project uses Jest, so the Mocha rules aren't applicable
- This project doesn't use React (yet), so the React rules aren't applicable
- `gulp-eslint` is pinned to ESLint 6.x, which is very out-of-date, and replaced here with manual Gulp handling (see https://github.com/adametry/gulp-eslint/issues/250).
- Babel rules trigger an error by absence of Babel configuration. Good idea to include anyways, so ESLint parsing is accurate.


zachmargolis (approved):
LGTM

PR #247, opened 2021-09-10
Define all variables as default values

aduth:
**Why**: To allow per-project customization, which previously required cumbersome workarounds.

Documentation: https://sass-lang.com/documentation/variables#default-values

Real-world examples: https://github.com/18F/identity-site/search?l=SCSS&q=todo


jmdembe (approved):
LGTM

PR #248, opened 2021-09-10
Remove width non-responsive override

aduth:
**Why**: These were kept in #236 as they differed from USWDS defaults, but they _were_ identical to the default as of the original file creation. USWDS later [changed the default](https://github.com/uswds/uswds/commit/16806bde246f61207cd64bfbf2e21c5a3a8b998b), and since we had no explicit reason to disable responsive styles, we should inherit the default instead.


zachmargolis (approved):
LGTM

PR #249, opened 2021-09-10
Remove uppercase transform from navbar styles

aduth:
**Why**: Per Figma design system "Layouts and templates". This is already applied to the brochure site via [style override](https://github.com/18F/identity-site/blob/25d5b444c3b8852b5000721837e84aef75a92512/_sass/components/_nav.scss#L191), and reflected in upcoming Partnerships site design.

Before|After
---|---
image|image


zachmargolis (approved):
LGTM

nickttng (approved):

PR #250, opened 2021-09-17
Fix typo in Rails asset helper documentation

aduth:
Previously: #170

**Why**: So that the documentation is accurate. Earlier iterations had named the helper method login_design_asset_path . Rename design_system_asset_path to improve future compatibility with naming and to be more clearly associated with design system.


mdiarra3 (approved):

orenyk (approved):
:shipit:

jmdembe (approved):
👍🏾

PR #251, opened 2021-09-17
Badge: Improve visual appearance and accessibility, simplify markup

aduth:
- Improved vertical centering
- Removed unused ID attribute
- Removed unnecessary `<span>` child to simplify markup and improve performance in deeply-nested markup documents
- Set null icon alt to treat as decorative / redundant, based on adjacent text providing the same information.
   - ["[...] the information provided by the image might already be given using adjacent text"](https://www.w3.org/WAI/tutorials/images/decorative/)
- Remove broken responsive styling

Before|After
---|---
image|image

### DIFF
--- a/src/scss/components/_verification-badge.scss
+++ b/src/scss/components/_verification-badge.scss
@@ -3,23 +3,12 @@
   background-color: map-get($site-palette, 'site-white');
   border: units(1px) solid map-get($site-palette, 'site-green');
   box-sizing: border-box;
-  display: inline-block;
+  display: inline-flex;
+  align-items: center;
   padding: units(1) units(2);
   white-space: nowrap;
 
   img {
     margin-right: units(1);
-  }
-
-  span {
-    vertical-align: text-top;
-  }
-}
-
-@media #{$theme-site-margins-breakpoint} {
-  .lg-verification-badge {
-    margin: units(-1) 0 0;
-    position: static;
-    text-align: right;
   }
 }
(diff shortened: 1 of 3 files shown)


src/scss/components/_verification-badge.scss
```diff
@@ -3,23 +3,12 @@
...
   img {
     margin-right: units(1);
   }
-
-  span {
-    vertical-align: text-top;
-  }
-}
-
-@media #{$theme-site-margins-breakpoint} {
```
aduth, line 19: `$theme-site-margins-breakpoint` comes from USWDS and is equal to the string value `"desktop"` ([source](https://github.com/uswds/uswds/blob/e1ed21783b089a2831504971b3426fbea1b5b752/src/stylesheets/settings/_settings-spacing.scss#L90)). `@media desktop {` is not a valid media query. This would have needed to use [the `at-media` mixin](https://github.com/uswds/uswds/blob/develop/src/stylesheets/core/mixins/_at-media.scss) to do anything. Based on the styles, I'm assuming this is meant to address some of the IdP-specific styling of verification badges as right-aligned on the account page at desktop viewports. Since it's project-specific, I don't think it's appropriate to keep.

mdiarra3 (approved):

aduth:
Test failure is for visual regression, and is expected based on how the icon placement has been adjusted slightly to improve vertical centering:

image

PR #254, opened 2021-09-29
Configure auto-updates for USWDS

aduth:
**Why**: To reduce confusion and maintenance overhead incurred by version drift, to simplify the sync process, and to take advantage of the latest features and bug fixes as quickly as they're made available.

Reference: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/enabling-and-disabling-version-updates#enabling-github-dependabot-version-updates

### DIFF
--- a/dependabot.yml
+++ b/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    allow:
+      - dependency-name: 'uswds'


zachmargolis (approved):
Oo interesting, would it make sense for us to have our various projects implement something similar for `identity-style-guide`?

aduth:
> Oo interesting, would it make sense for us to have our various projects implement something similar for `identity-style-guide`?

I like that idea! One problem with that at the moment (which also reinforces why it would be a good idea) is that we're very out-of-date for identity-style-guide in many of our projects, and there's quite a few breaking changes to sort through 😬  I suppose at worst it means pull requests that linger until we are (or Andrew is) able to deal with it.

PR #255, opened 2021-09-29
Update form validation error message styling

aduth:
**Why**: As an end-user, if I try to submit a form with invalid form values, I want to see consistent and accessible messaging that tells me what I need to complete before I can submit the form successfully.

References:

- LG-4235
- https://gsa-tts.slack.com/archives/C01JVKYE4KD/p1632934852000500

Before|After
---|---
image|image

Live preview URL: TBD

### DIFF
--- a/test/dist/scss.test.js
+++ b/test/dist/scss.test.js
@@ -29,12 +29,8 @@ test('if an asset-path function is defined, it is used to generate asset paths w
   const css = results.css.toString('utf-8');
 
   const [usaAlertWarning] = css.match(/\.usa-alert--warning:before\s?{[^}]+}/);
-  const [usaInputError] = css.match(
-    /\.usa-textarea\.usa-input--error:not\(\.usa-input--inline\)\s?{[^}]+}/,
-  );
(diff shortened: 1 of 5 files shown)


mdiarra3 (approved):

test/dist/scss.test.js
```diff
@@ -29,12 +29,8 @@ test('if an asset-path function is defined, it is used to generate asset paths w
   const css = results.css.toString('utf-8');
 
   const [usaAlertWarning] = css.match(/\.usa-alert--warning:before\s?{[^}]+}/);
-  const [usaInputError] = css.match(
-    /\.usa-textarea\.usa-input--error:not\(\.usa-input--inline\)\s?{[^}]+}/,
-  );
```
aduth, line 34: I removed this assertion because it seemed to be testing essentially the same thing as the preceding line, which is that images are rewritten using the `$image-path` variable.

nickttng (approved):
I tested w/ the Federalist build of the component. Looks like it's implemented as expected.

aduth:
The failing visual regression test is expected, since it reflects the larger border size and relocated icon.

PR #259, opened 2021-10-06
Backport additional search styles from brochure site

aduth:
- Full button text "Search" is always shown, regardless of viewport size
- Reduce padding for default search button from 2 units (1rem) to 1.5 units (0.75rem)
- Reduce font size for default search button from xs (1rem) to 2xs (0.87rem)

Live preview: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-backport-more-search/components/search/

### DIFF
--- a/docs/_components/search.md
+++ b/docs/_components/search.md
@@ -9,15 +9,16 @@ lead: >
   Search allows users to search for specific content if they know what search terms to use or can’t find desired content in the main navigation.
 ---
 
+{% include helpers/base-component.html component="search" %}
+
 ## Default
+
 Search should be easily found. For the default search version, consider displaying in the header and/or side navigation.
 
 {% capture example %}
-<form accept-charset="UTF-8" action="https://search.usa.gov/search" class="usa-search usa-search--smalldisplay-flex flex-justify-center tablet:grid-col-4" method="get" role="search">
-  <input name="utf8" type="hidden" value="&#x2713;"/>
-  <input name="affiliate" type="hidden" value="login.gov"/>
-  <label class="usa-sr-only" for="search-field-header-nav">Search</label>
-  <input class="usa-input" id="search-field-header-nav" name="query" type="search">
+<form class="usa-search" role="search">
+  <label class="usa-sr-only" for="tkBH">Search</label>
+  <input class="usa-input" id="tkBH" type="search">
   <button class="usa-button" type="submit">
     <span class="usa-search__submit-text">Search</span>
   </button>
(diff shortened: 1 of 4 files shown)


aduth:
The `2xs` font-size token produces a value of 0.87rem / 13.92px, which is very close to--but not identical to--the Figma designs' 0.875rem / 14px. I wouldn't have thought this would have much an impact, but it does seem noticeable to me in practice. While I'd generally prefer deferring to a standard design system token, we could make an exception here and hard-code this to `0.875rem` size?

`0.87rem`, Public Sans|`0.875rem`, Public Sans
---|---
image|image

nickttng:
Appreciate you noticing those kinds of details. To tie up some loose ends, it looks like the search component, both default and big variant, in the Figma library should be updated to reflect the LGDS, correct?

nickttng:
Yep, I'm able to help with that, and tag you when updates are made in Figma - To make sure that LGDS and the library are both aligned.

docs/_components/search.md
```diff
@@ -9,15 +9,16 @@ lead: >
...
 Search should be easily found. For the default search version, consider displaying in the header and/or side navigation.
 
 {% capture example %}
-<form accept-charset="UTF-8" action="https://search.usa.gov/search" class="usa-search usa-search--smalldisplay-flex flex-justify-center tablet:grid-col-4" method="get" role="search">
-  <input name="utf8" type="hidden" value="&#x2713;"/>
-  <input name="affiliate" type="hidden" value="login.gov"/>
-  <label class="usa-sr-only" for="search-field-header-nav">Search</label>
-  <input class="usa-input" id="search-field-header-nav" name="query" type="search">
+<form class="usa-search" role="search">
+  <label class="usa-sr-only" for="tkBH">Search</label>
```
zachmargolis, line 20: i'm not used to these randomized IDs, is that a new thing we're doing?

zachmargolis (approved):
LGTM

docs/_components/search.md
aduth: > i'm not used to these randomized IDs, is that a new thing we're doing?

It predates me, I've just kept it going 🤷  We would want unique IDs at least for each demo example, though I'm not sure the best way to include that in code snippets. USWDS goes the route of creating unique-but-meaningful IDs ([example](https://designsystem.digital.gov/components/text-input/)), which I think would work a bit better, though require more (up-front) effort. I've found these randomized IDs to be a bit of a pain actually, since I don't know that there's an easy way to enforce it, or even how best to create new ones.

zachmargolis: yeah the red code in the diff had the meaningful ones so that's why I was wondering why the random ones came back

aduth: > yeah the red code in the diff had the meaningful ones so that's why I was wondering why the random ones came back

Ah, I see now. I might still be in favor of the revisions, since:

- The IDs were copied verbatim from the brochure site. "heading" doesn't really apply in the context of this example.
- Consistency for consistency's sake

PR #270, opened 2021-11-17
Backport mobile navigation styles from brochure site

aduth:
**Why**: To be able to simplify and remove customizations from the brochure site, and to ensure consistent styling across all our products.

In order to demonstrate these changes, the pull request also includes revisions to show navigation submenu items in mobile navigation (e.g. individual "Components" sub-pages), which were not shown previously.

Live preview:

- Documentation: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-backport-mobile-nav/
- Brochure site: https://federalist-17bd62cc-77b7-4687-9c62-39b462ce6fd5.app.cloud.gov/preview/18f/identity-site/aduth-upgrade-design-system/

Corresponding brochure site changes: https://github.com/18F/identity-site/pull/723


nickttng:
@aduth Is this ready for review? If so, these screenshots below (before/after) is what I'm review, correct?


|Before|After|
|--:|---|
|lgds-mobile-nav-before|lgds-mobile-nav-after|

aduth:
@nickttng It is intended to be, yes, though I did have to infer some design to try to find a happy medium for different types of subnavigation behavior, and I'm definitely open for feedback to revisions you'd like for me to make. For that particular screen, it is meant to imitate the current brochure site Help Center selected article.

For example:

brochure site screenshot

As implemented here, it's not currently using the accordion menus, though the styles do exist and it could support it, which I'll hope to be able to demonstrate with a preview site for the brochure site using these changes.

For top-level pages, the "current page" styles are different, and should match the same behavior from brochure site:

Documentation Before|Documentation After|Brochure Site
---|---|---
design login gov_brand_(iPhone X)|federalist-2f194a10-945e-4413-be01-46ca6dae5358 app cloud gov_preview_18f_identity-style-guide_aduth-backport-mobile-nav_brand_(iPhone X)|login gov_what-is-login_(iPhone X)

Also, we didn't have submenu items previously, but this also brings those into alignment with the brochure site. As you can see on the IdP, these are currently quite small:

IdP submenu items

aduth:
@nickttng I had some trouble getting the preview branch to work for the [corresponding brochure site PR](https://github.com/18F/identity-site/pull/723), but I think it's working now.

You can test these changes live in the context of the brochure site here: https://federalist-17bd62cc-77b7-4687-9c62-39b462ce6fd5.app.cloud.gov/preview/18f/identity-site/aduth-upgrade-design-system/

Notably, this reflects the accordion submenu behavior in the "Help Center" section.

aduth:
@nickttng Would you be able to give this a look over?

nickttng:
I took a look at both LGDS site and brochure site previews.

## LGDS site

For Components, it looks like it has a discrepancy. It doesn't have the red vertical indicator left of "Components" like the "Utilities" example. Is there a reason why it does not appear?

|`/components/accordions/`|`/utilities/color/`|
|--:|--:|
|federalist-2f194a10-945e-4413-be01-46ca6dae5358 app cloud gov_preview_18f_identity-style-guide_aduth-backport-mobile-nav_components_accordions_(iPhone 6_7_8)|federalist-2f194a10-945e-4413-be01-46ca6dae5358 app cloud gov_preview_18f_identity-style-guide_aduth-backport-mobile-nav_utilities_color_(iPhone 6_7_8)|

## Brochure site

Overall, it functions and looks as expected; however, do we want to address when to use "+/-", red vertical bar indicator, etc? It looks like a discrepancy to remedy.

## Account management

Font-wise, the submenu items do look small. Also, does this menu displays all submenu items by default rather than hidden?

aduth:
Thanks for taking a look @nickttng !

>For Components, it looks like it has a discrepancy. It doesn't have the red vertical indicator left of "Components" like the "Utilities" example. Is there a reason why it does not appear?

Good catch. I think this is mostly an implementation inconsistency here in the documentation site, a matter of adding or leaving out the `usa-current` class on the parent page. Do you have a preference one way or the other whether it should be visible?

I think this ties to your later point about 'when to use "+/-", red vertical bar indicator'. For context, I believe we haven't shown the red bar in the brochure site when viewing a help center article.

>Font-wise, the submenu items do look small.

In the current implementation, I'm overriding them to be smaller. If you think it's too small, I could just remove that and let it be shown at the default size?

I can make that change and we can see how it feels.

>Also, does this menu displays all submenu items by default rather than hidden?

It should only default to show items when viewing a page within that section. This is also an implementation detail of the documentation site, so we could change it here, or revise it for other projects depending how we think makes most sense.

PR #273, opened 2021-11-17
Fix line wrapping for Buttons documentation

aduth:
**Why**: So that all of the button previews appear on the same line, and for consistency with how this is already handled for larger button variants.

By default, content in a Markdown file is rendered in paragraph elements within a `usa-prose` block which constrains the width of the content. This is a good thing for text content, since it helps limit line length. However, it doesn't make sense for component previews.

**Screenshots:**

Before|After
---|---
image|image

Review with whitespace hidden: https://github.com/18F/identity-style-guide/pull/273/files?w=1

### DIFF
--- a/docs/_components/buttons.md
+++ b/docs/_components/buttons.md
@@ -21,11 +21,13 @@ subnav:
 <button class="usa-button">
 ```
 
-<button class="usa-button">Default</button>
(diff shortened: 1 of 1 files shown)


zachmargolis (approved):
LGTM

aduth:
Visual regression "failure" is for the expected improvement:

image

PR #271, opened 2021-11-17
Add "Coding Standards" section to contributing guide

aduth:
**Why**: To better communicate the approach to extending USWDS.

PR #275, opened 2021-11-22
Match usa-input-error selector to USWDS

aduth:
**Why**: The [error input style selector in USWDS](https://github.com/uswds/uswds/blob/eb84a497042aa85d79e8b5d2225a874c210bdb33/src/stylesheets/elements/form-controls/_text-input.scss#L11-L13) does not qualify as one of `.usa-input`, `.usa-textarea`, or `.usa-select`. In many cases, this makes little difference since the modifier would be applied to an element with one of those classes. However, this is not strictly required, and causes some [challenge in the IdP with SimpleForm compatibility](https://github.com/18F/identity-idp/pull/5626), where design system classes are not yet being applied (pending LG-3877). Furthermore, per updated guidance in #271, selectors with overriding styles should seek to precisely match the USWDS selector it is overriding.

Future work should seek to bring stylesheet files into closer alignment with USWDS files where these styles are defined (tracked in LG-5398). This would include moving styles from `src/scss/components/_inputs.scss` to `src/scss/elements/form-controls/_text-input.scss`.

**Screenshots:**

Before|After
---|---
Screen Shot 2021-11-22 at 9 29 37 AM|Screen Shot 2021-11-22 at 9 29 43 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@
 
 - Upgraded USWDS from v2.12.1 to v2.12.2 (see [release notes](https://github.com/uswds/uswds/releases/tag/v2.12.2)) ([#269](https://github.com/18F/identity-style-guide/pull/269))
 
+### Bug Fixes
+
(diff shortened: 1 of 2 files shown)


zachmargolis (approved):
LGTM

PR #276, opened 2021-12-09
Use login.gov shared Stylelint configuration

aduth:
Related: https://github.com/18F/identity-idp/pull/5689

**Why:** So that we can use a common standard ruleset across all repositories where Sass is used.


gulpfile.js
```diff
@@ -9,7 +9,7 @@ const replace = require('gulp-replace');
 const rename = require('gulp-rename');
 const gulpif = require('gulp-if');
 const sass = require('gulp-sass')(require('sass'));
-const gulpStylelint = require('gulp-stylelint');
```
aduth, line 12: `gulp-stylelint` may be abandoned, as it doesn't support the newest major release of Stylelint released early in the year, and the maintainer hasn't been active in related issues (https://github.com/olegskl/gulp-stylelint/issues/132).

In any case, the cumulative dependency shifting here removed _a lot_ from the dependency tree (net -1859 lines in `package-lock.json`), which I'm pretty happy about.

mitchellhenke (approved):
👍🏼 👍🏼 👍🏼 👍🏼

PR #278, opened 2022-01-14
Use sass-embedded (native Dart) SASS compiler

aduth:
**Why**: Because it's much faster, and to provide improved guarantees of Dart SASS compatibility by using it internally.

**Comparison:**

_Before:_

```
$ npm exec gulp build-sass
...
[15:15:48] Finished 'build-sass' after 32 s
```

_After:_

```
$ npm exec gulp build-sass
...
[15:14:01] Finished 'build-sass' after 4.84 s
```

**Context:**

The `sass` NPM package is a version of the Dart SASS source, but compiled to JavaScript. This JavaScript compilation comes at the cost of performance, which is clearly documented in the [SASS installation instructions](https://sass-lang.com/install):

>However, please note that this will install the pure JavaScript implementation of Sass, which runs somewhat slower than the other options listed here.

The [`sass-embedded` package](https://www.npmjs.com/package/sass-embedded) is an official-but-"beta" version of the SASS compiler which automatically downloads OS-specific binaries to run the native Dart executable, resulting in much improved performance.

**Implementation notes:**

The inclusion of the `luminance` function override serves two purposes:

1. To fix an infinite loop which occurs only in Dart SASS, explained at https://github.com/uswds/uswds/pull/4437#issuecomment-1013428155
2. Improved performance

**Next:**

- The majority of the `build-sass` task is now occupied by the PostCSS minification + autoprefixing (removing PostCSS processing yields build times of ~2.0s). [Parcel's new Rust-based CSS transformer](https://github.com/parcel-bundler/parcel-css) could be a good substitute here, since it serves these two requirements quite well.
- Aiming for a similar set of changes in the IdP to resolve issues with Sprockets + sassc-rails


src/scss/packages/_required.scss
```diff
@@ -4,7 +4,21 @@
 @import '../uswds-theme/color';
 @import '../uswds-theme/utilities';
 @import '../uswds-theme/components';
-@import '../uswds/packages/required';
```
aduth, line 7: The changes here expand out the contents of [the `_required.scss` package](https://github.com/uswds/uswds/blob/develop/src/stylesheets/packages/_required.scss) to be able to inject the override immediately following [`_functions.scss`](https://github.com/uswds/uswds/blob/823b7a69af6fecce27b0c1de9dc950c146f4ddf4/src/stylesheets/packages/_required.scss#L12), because the luminance function is used later int he same `_required.scss` through [`placeholders/_all.scss`](https://github.com/uswds/uswds/blob/823b7a69af6fecce27b0c1de9dc950c146f4ddf4/src/stylesheets/packages/_required.scss#L17).

test/dist/scss.test.js
```diff
@@ -1,35 +0,0 @@
-const path = require('path');
-const { renderSync } = require('sass');
```
aduth, line 2: In retrospect, I probably could have updated this to use the `sass-embedded` package, but this test has been a source of some headaches (mostly slowness failing tests), and I'm not sure how much value it provides.

zachmargolis (approved):
LGTM

PR #279, opened 2022-01-14
LG-4791: Add Process List component styles

aduth:
Base styling is smaller and excludes the connector otherwise shown with the base [USWDS Process List component](https://designsystem.digital.gov/components/process-list/).

A "connected" (`.usa-process-list--connected`) variant restores the connector.

A "big" (`.usa-process-list--big`) variant increases the size of the numbered circle.

They can also be used in combination with the other.

Live preview URL: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-process-list/components/process-list/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### New Features
 
 - Field validation success is added to form validation ([#265](https://github.com/18F/identity-style-guide/pull/265))
+- Added custom styling for the Process List component. ([#279](https://github.com/18F/identity-style-guide/pull/279))
 
 ### Improvements
 
--- a/docs/_components/process-list.md
+++ b/docs/_components/process-list.md
@@ -1,0 +1,49 @@
+---
+title: Process List
+lead: A process list displays the steps or stages of important instructions or processes.
+---
+
+{% include helpers/base-component.html component="process-list" stylesheet="process-list" %}
+
+{% capture example %}
+<ol class="usa-process-list">
+  <li class="usa-process-list__item">
+    <h4 class="usa-process-list__heading">Start a process</h4>
+    <p class="margin-top-05">
+      Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Morbi commodo,
+      ipsum sed pharetra gravida, orci magna rhoncus neque.
+    </p>
+    <ul>
+      <li>
+        Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Morbi commodo,
+        ipsum sed pharetra gravida, orci magna rhoncus neque, id pulvinar odio
+        lorem non turpis.
+      </li>
+      <li>
+        Nullam sit amet enim. Suspendisse id velit vitae ligula volutpat
+        condimentum.
+      </li>
+      <li>Aliquam erat volutpat. Sed quis velit.</li>
+    </ul>
+  </li>
+  <li class="usa-process-list__item">
+    <h4 class="usa-process-list__heading">Proceed to the second step</h4>
+    <p>
+      Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Morbi commodo,
+      ipsum sed pharetra gravida, orci magna rhoncus neque, id pulvinar odio
+      lorem non turpis. Nullam sit amet enim. Suspendisse id velit vitae ligula
+      volutpat condimentum. Aliquam erat volutpat. Sed quis velit. Nulla
+      facilisi. Nulla libero. Vivamus pharetra posuere sapien.
+    </p>
+  </li>
+  <li class="usa-process-list__item">
+    <h4 class="usa-process-list__heading">Complete the step-by-step process</h4>
+    <p>
+      Nullam sit amet enim. Suspendisse id velit vitae ligula volutpat
+      condimentum. Aliquam erat volutpat. Sed quis velit. Nulla facilisi. Nulla
+      libero. Vivamus pharetra posuere sapien.
+    </p>
+  </li>
+</ol>
+{% endcapture %}
+{% include helpers/code-example.html code=example %}
--- a/docs/_components/process-list.md
+++ b/docs/_components/process-list.md
@@ -5,8 +5,55 @@ lead: A process list displays the steps or stages of important instructions or p
 
 {% include helpers/base-component.html component="process-list" stylesheet="process-list" %}
 
+## Default
+
 {% capture example %}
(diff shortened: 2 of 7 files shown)


aduth:
Visual regression build failure is because we're adding new content, so the changed navigation is flagged.

cdelivuk:
@aduth  i like the look of the process list. comparing it to the one on the get started page, can you confirm the color of the vertical connectors? yours looks a hair darker, which i like. obviously once implemented we can swap whatever is being used on the site for the design system. i also like the two sizes. lgtm

aduth:
> comparing it to the one on the get started page, can you confirm the color of the vertical connectors?

It's using the `base-lighter` token (`#dedede`), which I believe is the same as what's in the Partner Site designs. Is that right?

nickttng:
Overall, it looks to be in the right direction.

Regarding the default and connected lists: While the white item number on `primary` / `#0071bb` has the contrast ratio of `5.14:1`, it does not pass the WCAG AAA with "Normal Text." 

Any thoughts/concerns? Should we consider darkening the background to `primary-dark` / `#205493`? A potential downside with the darkened background is that the item number might not stand out as much among other elements (headings, body text, etc.)

nickttng:
What's the possibility of a team member reusing the code snippet with the default `h4` but hasn't tailored the correct heading level in their own implementation? Thinking about what might be easily overlooked and if we should make it explicit on the component page instead of relying on the USWDS link for more guidance.

cdelivuk:
@nickttng great point. I didn't think about AAA contrast on a smaller circle/number. here is a screenshot using primary-dark / #205493 for comparison.
Screen Shot 2022-01-20 at 9 08 21 AM

cdelivuk:
> It's using the `base-lighter` token (`#dedede`), which I believe is the same as what's in the Partner Site designs. Is that right?

yep- that is correct. it might be my ole eyes 😂

aduth:
>Regarding the default and connected lists: While the white item number on primary / #0071bb has the contrast ratio of 5.14:1, it might not pass the WCAG AAA with "Normal Text." (Depending on the font size) Any thoughts/concerns?

Happy to make a change here. The numbers are bolded (700 weight) and 1rem (16px) for the default styling, 1.5rem (24px) for the large size, so passing WCAG 2 AA minimums for both with current default and large sizing, and failing WCAG 2 AAA minimum for default size, still passing for large size. Based on the WCAG 3 tool you linked, it would be a Score 5 (max) for both.

Based on that, I might say it's fine as-is if we'd want? But more contrast couldn't hurt either.

>What's the possibility of a team member reusing the code snippet with the default h4 but hasn't tailored the correct heading level in their own implementation? Thinking about what might be easily overlooked and if we should make it explicit on the component page instead of relying on the USWDS link for more guidance.

The code snippets were copied verbatim from the USWDS documentation, but yes, it would need to be adapted to the specific implementation to make sure the heading levels increment in the correct order. This is also something I expect would (should) be caught by automated tests as well. The font sizes are explicitly assigned, so it should appear visually identical regardless which level is used.

That being said, the USWDS documentation does include [an extra snippet of advice to this point](https://designsystem.digital.gov/components/process-list/#accessibility-process-list):

>**Use semantic heading levels.** Though our default code uses an `<h4>`, use the correct heading level with the class name `usa-process-list__heading` in your own implementation.

Maybe we could include a short preamble warning about this, similar to the note we include in the [Accordion documentation](https://design.login.gov/components/accordions/)? Then again, it feels a bit strange to cherry-pick which guidance to include, vs. featuring the base documentation more prominently or copying all of it into our documentation.

nickttng:
> But more contrast couldn't hurt either.

Let's update to `primary-dark`.

> Then again, it feels a bit strange to cherry-pick which guidance to include, vs. featuring the base documentation more prominently or copying all of it into our documentation.

Good point. I flagged about the semantic headings because historically, it has been an issue that's overlooked in the implementation. 

Alternatively, I wonder if the language of the "View USWDS component" URL could be more clear, like "View USWDS component for guidance" (adding "for guidance").

aduth:
> > But more contrast couldn't hurt either.
> 
> Let's update to `primary-dark`.

Cool, done in ce0f5a6.

> Alternatively, I wonder if the language of the "View USWDS component" URL could be more clear, like "View USWDS component for guidance" (adding "for guidance").

I like that idea. I added it in f392e10.

Both of these changes should be reflected in the live preview link shortly.

nickttng:
Not critical, but something to note: The vertical center for the item number appears slightly more upward when my browser (both Chrome and Safari) is at 100% (See screenshot). When I resize to either 90% or 110%, the vertical center looks fine. Does that happen on your side too?

|90%|100%|110%|
|--:|--:|---|
|Screen Shot 2022-01-20 at 3 45 08 PM|Screen Shot 2022-01-20 at 3 44 55 PM|Screen Shot 2022-01-20 at 3 45 02 PM|

aduth:
> Not critical, but something to note: The vertical center for the item number appears slightly more upward when my browser (both Chrome and Safari) is at 100% (See screenshot).

Yeah, I did notice this. I had actually put some "nudge" styling in place originally, then decided to revert it in 20b09bf. I noticed that it differs depending on the resolution of the display, because the nudge caused an offset when I was looking at the preview on my phone. Since the default styling is actually centered, I think it's probably a sub-pixel rounding issue.

There might be some other adjustments we could try to make this a bit more consistent (e.g. font size, circle size, etc).

aduth:
I narrowed the connector slightly (8px to 5px) for the default (non-"big") process list based on [this discussion in Slack](https://gsa-tts.slack.com/archives/C01JVKYE4KD/p1642783068003800) (cc @anniehirshman-gsa), which can be seen in the [live preview](https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-process-list/components/process-list/).

Are there any other revisions we'd like to make here?

nickttng:
No other revisions. Do we want to figure out the vertical center issue in this PR? Or follow up another time?

nickttng:
This is quite a rabbit hole. 🕳️  

Appreciate you for investigating. Except for the potential future CSS support, I don't think any is worth the workaround for something that is minor and browser-based. Let's leave it for another time. 👍🏼

nickttng (approved):

solipet (approved):
🚀

PR #281, opened 2022-01-20
Set font-size of tile checkbox/radio to 1.125rem

aduth:
**Why**: To avoid font-face-specific sizing from causing the label text from being too small relative to the accompanying description.

**Screenshots:**

Before|After
---|---
before|after

Live preview URL: TBD


anniehirshman-gsa (approved):
LGTM

PR #283, opened 2022-01-24
Replace "serve" dependency with Ruby built-in static file server

aduth:
**Why**:

Fewer dependencies means:

- Faster install times
- Less maintenance overhead
- Smaller vulnerability surface area
- +2, -1255 changes

### DIFF
--- a/jest-puppeteer.config.js
+++ b/jest-puppeteer.config.js
@@ -1,6 +1,6 @@
 module.exports = {
   server: {
-    command: `./node_modules/.bin/serve --listen ${process.env.JEST_PORT} --no-clipboard ./dist`,
+    command: `python -m http.server ${process.env.JEST_PORT} --directory ./dist`,
     port: parseInt(process.env.JEST_PORT, 10),
   },
 };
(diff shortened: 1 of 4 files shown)


jest-puppeteer.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   server: {
-    command: `./node_modules/.bin/serve --listen ${process.env.JEST_PORT} --no-clipboard ./dist`,
+    command: `ruby -run -ehttpd ./dist -p${process.env.JEST_PORT}`,
```
zachmargolis, line 3: TIL! I had been doing `python -m SimpleHTTPServer` but it's nice to keep everything to one language

zachmargolis (approved):
LGTM

jest-puppeteer.config.js
aduth: > TIL! I had been doing `python -m SimpleHTTPServer` but it's nice to keep everything to one language

I actually used the Python version initially (07f1f5e) but:

- The command differs between Python 2 and Python 3
- Python isn't available in the CircleCI Ruby images, which makes sense in retrospect

aduth: Also, of interest: https://gist.github.com/willurd/5720255

PR #284, opened 2022-01-24
Set checkbox/radio label line-height to match body

aduth:
**Why**: So that multi-line checkbox or radio fields do not appear tighter relative to surrounding body copy.

Note: This also changes the default line-height, configured to match secure.login.gov (changes from 1.6 default to 1.5).

Before|After
---|---
before|after


zachmargolis (approved):
LGTM

nickttng (approved):
Looks good to me. For your awareness,I added a self-reminder to check line-height for multiple lines in the Figma library.

PR #285, opened 2022-01-25
Replace pa11y with underlying HTML_CodeSniffer

aduth:
**Why**:

- It currently duplicates and sometimes conflicts with our integration testing setup
   - Currently, we run two separate versions of Puppeteer, which each download separate copies of Chromium
   - Also helps me to avoid debugging a vague error occurring when upgrading dependencies to latest version
- Consolidates testing tools to Jest
- A step toward alignment of accessibility integration testing, as it could be reasonably substituted with aXe (our runner of choice in 18F/identity-site and 18F/identity-idp).
- Easier to run individual tests, which was difficult before since Jest setup occurred through scripts/jest.sh and did not support additional arguments. Now, test commands can go directly through the Jest CLI.
- Fewer dependencies overall (net 1636 lines removed from `package-lock.json`)

Note that the intended result of these changes is an identical testing result to what existed previously, since `pa11y-ci` was already running HTML_CodeSniffer on our behalf. For additional context on HTML_CodeSniffer vs. aXe, see #203.


package.json
```diff
@@ -34,8 +34,6 @@
     "prelint": "make build-package",
     "lint": "gulp lint",
     "test": "make test",
-    "test-pa11y": "make test-runner-pa11y",
-    "test-jest": "make test-runner-jest",
     "build": "make build",
```
zachmargolis, line 37: I guess I've never looked inside this file, but it seems very weird to me that the npm tasks call out to the Makefile... seems like most of the pipeline happens in JS except the jekyll build... make in a separate/future PR we can invert this?

zachmargolis (approved):
LGTM

package.json
aduth: > I guess I've never looked inside this file, but it seems very weird to me that the npm tasks call out to the Makefile... seems like most of the pipeline happens in JS except the jekyll build... make in a separate/future PR we can invert this?

Yeah, it's a bit nonconventional as things stand. Also means there is no `make run` like there is in other projects.

I feel like `Makefile` and npm scripts serve many of the same goals, so maybe we could just get rid of one or the other. 🤷

PR #290, opened 2022-02-04
Remove vertical padding from Process List component

aduth:
**Why**: The Process List component includes some top and bottom padding inherited from USWDS, making it difficult to position in content, where surrounding content's own margins may cause excess space. Most components don't (and shouldn't) include any inherent vertical spacing, allowing the developer to decide to apply margins as appropriate, or rely on margins from surrounding content. Margins also have the added benefit of allowing for [margin collapse](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) behavior, to allow the component to be more easily reused in content without awareness of the specific context of its surrounding content.

Related discussion: https://github.com/18F/identity-idp/pull/5863#issuecomment-1028374791

Live preview URL: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-rm-process-list-padding/components/process-list/

**Implementation notes:**

Since the "Big" variant circle can otherwise appear outside of the bounds of the Process List if no padding were applied, this implementation includes enough padding that the Process List bounds begin at the far extent of the first item's circle:

image

This padding does not exist for the default size, since the circle should appear at an identical size to the heading text:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Improvements
+
+- The Process List component no longer applies vertical padding which would affect its layout relative to surrounding content. ([#290](https://github.com/18F/identity-style-guide/pull/290))
+
 ## 6.3.1
 
 ### Bug Fixes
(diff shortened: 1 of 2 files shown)


nickttng (approved):
Looks great to me. Compared between the existing and preview versions, and I appreciate how much you paid attention to those details.

PR #291, opened 2022-02-07
Update line-height behavior to respect token value, size headings relative to body text

aduth:
**Why**:

- So that assigned line-heights match expectations based upon [the available token options](https://github.com/uswds/uswds/blob/61a0d99f0e6b36c3758948ba6ac46140abc5e585/src/stylesheets/theme/_uswds-theme-typography.scss#L363-L371), not influenced by either font family or scale
- So that heading line-heights can be smaller than body content, but only in the case that the actual value would be greater than that of body text (and in all other cases, using the body line-height for headings)

Note that this diverges slightly from original decisions to use a line-height value of 1.33 for headings. USWDS includes a line-height token for 1.35. Previously, due to normalization, this value would never take effect. However, with this pull request also removing the normalization behavior, the 1.35 actual value for headings can be used. Since this is very close to 1.33, and avoids circumventing USWDS tokens, would we be okay with this value, or should we force it to 1.33?

Live preview URL: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-normalize-heading-line-heights/utilities/typography/

Effective difference in line-height:

Heading|Before|After
---|---|---
h1|1.4|1.35
h2|1.4|1.35
h3|1.4|1.5
h4|1.4|1.5
h5|1.4|1.5
h6|1.4|1.5

**Screenshots:**

Before|After
---|---
design login gov_utilities_typography_|localhost_4000_utilities_typography_


nickttng:
Okay w/ `1.35` as the value for both H1 and H2. 

@anniehirshman-gsa Any preference?

anniehirshman-gsa (approved):
Line height is looking good! And I agree, 1.35 it is

nickttng (approved):

PR #292, opened 2022-02-08
Change body font size to USWDS default "sm"

aduth:
Related Slack discussion: https://gsa-tts.slack.com/archives/CNCGEHG1G/p1644337817896069?thread_ts=1644246272.507329&cid=CNCGEHG1G

**Why**: Previously with Source Sans, a body font size of "xs" would result in an effective font size of 1rem, due to scaling resulting from its [smaller cap height](https://github.com/uswds/uswds/blob/61a0d99f0e6b36c3758948ba6ac46140abc5e585/src/stylesheets/core/_system-tokens.scss#L282) relative to the [base cap height](https://github.com/uswds/uswds/blob/61a0d99f0e6b36c3758948ba6ac46140abc5e585/src/stylesheets/core/_system-tokens.scss#L416). In Public Sans, "xs" results in an effective font size of 0.94rem. To restore the desired 1rem body font size, we should drop this override and inherit the USWDS default "sm".

Live preview URL: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-default-font-size/utilities/typography/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@
 
 ### Improvements
 
-- Line height calculations are improved such that the [desired token size](https://github.com/uswds/uswds/blob/61a0d99f0e6b36c3758948ba6ac46140abc5e585/src/stylesheets/theme/_uswds-theme-typography.scss#L363-L371) will always apply regardless of font family or scale.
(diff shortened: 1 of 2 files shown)


anniehirshman-gsa (approved):
LGTM x2!

PR #293, opened 2022-02-09
Drop cssnano in favor of Sass built-in compression

aduth:
**Why**:

- Fewer dependencies
- Faster build

```
NODE_ENV=production npm exec gulp build-sass
```

Before|After
---|---
`[10:13:02] Finished 'build-sass' after 6.55 s`|`[10:11:57] Finished 'build-sass' after 2.19 s`


zachmargolis (approved):
LGTM

PR #296, opened 2022-02-15
Add illustration to radio and checkbox tiles

jmdembe:
This PR adds an illustration to radio and checkbox tiles.

**Why**: To facilitate development of MFA checkboxes and radio labels


jmdembe:
Acknowledging that visual regression test will fail as we are adding a new checkbox and radio button tile.

src/scss/components/_inputs.scss
```diff
@@ -144,6 +144,27 @@ $input-select-margin-right: 1;
   }
 }
 
+.tile-container {
+  @include at-media('tablet') {
+    display: flex;
+  }
+  &__image {
+    width: 1em;
```
aduth, line 152: I think we'll want to use `rem` units here and throughout so that doesn't scale to the container text size.

src/scss/components/_inputs.scss
```diff
@@ -144,6 +144,27 @@ $input-select-margin-right: 1;
   }
 }
 
+.tile-container {
+  @include at-media('tablet') {
+    display: flex;
+  }
+  &__image {
```
aduth, line 151: Minor, personal preference: I find that this sort of nesting makes it difficult to backtrack from a class name to its associated styles, since searching the project for "tile-container__image" would not produce this as a result. I'd rather see the name spelled out in full.

src/img/fingerprint-icon.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="42" height="48" viewBox="0 0 42 48" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
jmdembe: I have optimized this file using SVGOMG. I don't see the check in the build, so I think it will be a good idea to add it in the future.

src/img/fingerprint-icon.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 42 48">
```
aduth, line 1: Maybe a broader discussion than we need to solve here, but I wonder if we should exclude these sorts of images from the design system package, since they're pretty use-case-specific. We have a few already though, so I think it'd be fine to follow the pattern.

src/scss/components/_inputs.scss
```diff
@@ -144,6 +144,30 @@ $input-select-margin-right: 1;
   }
 }
 
+.usa-checkbox__label-illustrated,
+.usa-radio__label-illustrated {
+  @include at-media('tablet') {
+    display: flex;
+  }
+  & .usa-radio__image,
```
aduth, line 152: It's safe to leave out the `&`, since it means the same thing anyways.

```suggestion
  .usa-radio__image,
```

src/scss/components/_inputs.scss
```diff
@@ -144,6 +144,30 @@ $input-select-margin-right: 1;
...
+    width: 1rem;
+    float: left;
+    margin-right: 0.5rem;
+    @include at-media('tablet') {
+      width: 2rem;
+      object-fit: contain;
+      align-self: flex-start;
+    }
+  }
+  & .usa-radio__label-text,
```
aduth, line 163: Same note on omitting the `&`:

```suggestion
  .usa-radio__label-text,
```

src/scss/components/_inputs.scss
```diff
@@ -144,6 +144,30 @@ $input-select-margin-right: 1;
   }
 }
 
+.usa-checkbox__label-illustrated,
+.usa-radio__label-illustrated {
```
aduth, line 148: There should be two dashes for the modifier, since we follow [BEM two-dashes style](https://en.bem.info/methodology/naming-convention/#two-dashes-style).

```suggestion
.usa-checkbox__label--illustrated,
.usa-radio__label--illustrated {
```

docs/_components/form-fields.md
```diff
@@ -159,6 +159,16 @@ Three styles of radio buttons are provided.
       <input id="a8c1" class="usa-radio__input usa-radio__input--tile" type="radio" name="a8c1">
       <label for="a8c1" class="usa-radio__label">Option - Unselected<span class="usa-radio__label-description">Support text</span></label>
     </li>
+    <li>
+      <input id="abc123" class="usa-radio__input usa-radio__input--tile" type="radio" name="abc123">
+      <label for="abc123" class="usa-radio__label usa-radio__label-illustrated">
+        <img class="usa-radio__image" src="{{ site.baseurl}}/assets/img/fingerprint-icon.svg">
+        <div class="usa-radio__label-text">
+          <span>Option - Illustration</span>
```
aduth, line 167: It would be nice if we could avoid this extra span. Is it needed?

docs/_components/form-fields.md
```diff
@@ -159,6 +159,18 @@ Three styles of radio buttons are provided.
       <input id="a8c1" class="usa-radio__input usa-radio__input--tile" type="radio" name="a8c1">
       <label for="a8c1" class="usa-radio__label">Option - Unselected<span class="usa-radio__label-description">Support text</span></label>
     </li>
+    <li>
+      <input id="abc123" class="usa-radio__input usa-radio__input--tile" type="radio" name="abc123">
+      <label for="abc123" class="usa-radio__label">
+        <div class="container">
+          <img class="container__image" src="{{ site.baseurl}}/assets/img/fingerprint-icon.svg">
```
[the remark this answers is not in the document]
jmdembe: We are considering this to be decorative, so `alt=""` will be good here

docs/_components/form-fields.md
jmdembe: It was not needed, so I removed it.

src/img/fingerprint-icon.svg
jmdembe: I think it would be good to exclude because it would help the package size.

docs/_components/form-fields.md
```diff
@@ -159,6 +159,18 @@ Three styles of radio buttons are provided.
       <input id="a8c1" class="usa-radio__input usa-radio__input--tile" type="radio" name="a8c1">
       <label for="a8c1" class="usa-radio__label">Option - Unselected<span class="usa-radio__label-description">Support text</span></label>
     </li>
+    <li>
+      <input id="abc123" class="usa-radio__input usa-radio__input--tile" type="radio" name="abc123">
+      <label for="abc123" class="usa-radio__label usa-radio__label--illustrated">
+      <div class="usa-radio__image">
+        <img alt="" src="{{ site.baseurl}}/assets/img/fingerprint-icon.svg">
+      </div>
```
aduth, line 167: Nit: Indentation

```suggestion
        <div class="usa-radio__image">
          <img alt="" src="{{ site.baseurl}}/assets/img/fingerprint-icon.svg">
        </div>
```

docs/_components/form-fields.md
```diff
@@ -256,6 +268,18 @@ Three styles of checkboxes are provided as well.
       <input id="c288" class="usa-checkbox__input usa-checkbox__input--tile" type="checkbox" name="c288">
       <label for="c288" class="usa-checkbox__label">Option - Unchecked<span class="usa-checkbox__label-description">Support text</span></label>
     </li>
+    <li>
+      <input id="abc123" class="usa-checkbox__input usa-checkbox__input--tile" type="checkbox" name="abc123">
+        <label for="abc123" class="usa-checkbox__label usa-checkbox__label--illustrated">
```
aduth, line 273: Nit: Indentation

```suggestion
      <label for="abc123" class="usa-checkbox__label usa-checkbox__label--illustrated">
```

aduth, line 273: Can we make these IDs unique from the previous example, since I expect there may be accessibility errors with how we have two sets of `abc123` inputs/labels.

src/img/fingerprint-icon.svg
aduth: If we wanted to do it as part of this pull request, we could put the file somewhere in `docs`, like `docs/img/fingerprint-icon.svg`, then reference it in the documentation as `/img/fingerprint-icon.svg`.

src/scss/components/_inputs.scss
```diff
@@ -144,6 +144,29 @@ $input-select-margin-right: 1;
...
+    @include at-media('tablet') {
+      width: 2rem;
+    }
+  }
+  .usa-radio__label--text,
+  .usa-checkbox__label--text {
+    @include at-media('tablet') {
+      flex: 1;
+    }
+  }
```
aduth, line 167: What are these styles for?

src/scss/components/_inputs.scss
```diff
@@ -144,6 +144,29 @@ $input-select-margin-right: 1;
...
+.usa-checkbox__input--tile + .usa-checkbox__label--illustrated,
+.usa-radio__input--tile + .usa-radio__label--illustrated {
+  @include at-media('tablet') {
+    display: flex;
+  }
+
+  .usa-radio__image,
+  .usa-checkbox__image {
+    width: 1rem;
+    float: left;
```
aduth, line 156: Does this need to be floated if we're also using flexbox?

aduth (approved):
Looking good 👍  Could you also include a note in the CHANGELOG.md file?

src/scss/components/_inputs.scss
jmdembe: This needs to be floated because on mobile, we are not using flexbox. On mobile, the design has the text wrapped around the image. 
image

aduth: Ah, good point, I missed that! Though, that raises another question: If the heading spans multiple lines, should it wrap under the icon (float behavior), or should the title text always be aligned with itself?

image

src/scss/components/_inputs.scss
jmdembe: This is to keep the sizing of the image when there is a lot of text in the label. I found that if there are multiple lines for either the title or the description text, the image would shrink accordingly.

src/scss/components/_inputs.scss
jmdembe: yes, this is/was a problem. I took the image out of the container and applied the styles to the image itself. This is now what we are working with
image

docs/_components/form-fields.md
```diff
@@ -159,6 +159,16 @@ Three styles of radio buttons are provided.
       <input id="a8c1" class="usa-radio__input usa-radio__input--tile" type="radio" name="a8c1">
       <label for="a8c1" class="usa-radio__label">Option - Unselected<span class="usa-radio__label-description">Support text</span></label>
     </li>
+    <li>
+      <input id="g6pz" class="usa-radio__input usa-radio__input--tile" type="radio" name="g6pz">
+      <label for="g6pz" class="usa-radio__label usa-radio__label--illustrated">
+        <img alt="" class="usa-radio__image" src="/img/fingerprint-icon.svg">
```
aduth, line 165: For Federalist compatibility, we might want to prefix with base URL.

```suggestion
        <img alt="" class="usa-radio__image" src="{{ site.baseurl }}/img/fingerprint-icon.svg">
```

docs/_components/form-fields.md
```diff
@@ -256,6 +266,18 @@ Three styles of checkboxes are provided as well.
       <input id="c288" class="usa-checkbox__input usa-checkbox__input--tile" type="checkbox" name="c288">
       <label for="c288" class="usa-checkbox__label">Option - Unchecked<span class="usa-checkbox__label-description">Support text</span></label>
     </li>
+    <li>
+      <input id="hx88" class="usa-checkbox__input usa-checkbox__input--tile" type="checkbox" name="hx88">
+        <label for="hx88" class="usa-checkbox__label usa-checkbox__label--illustrated">
+        <div class="usa-checkbox__image">
+          <img alt="" src="/img/fingerprint-icon.svg">
```
aduth, line 273: Same:

```suggestion
          <img alt="" src="{{ site.baseurl }}/img/fingerprint-icon.svg">
```

docs/_components/form-fields.md
```diff
@@ -256,6 +266,18 @@ Three styles of checkboxes are provided as well.
...
+    <li>
+      <input id="hx88" class="usa-checkbox__input usa-checkbox__input--tile" type="checkbox" name="hx88">
+        <label for="hx88" class="usa-checkbox__label usa-checkbox__label--illustrated">
+        <div class="usa-checkbox__image">
+          <img alt="" src="/img/fingerprint-icon.svg">        
+        </div>
+        <div class="usa-checkbox__label--text">
+          Option - Illustration
+          <span class="usa-checkbox__label-description">Support text</span>
+        </div>
```
aduth, line 278: Nit: Excess space.

```suggestion
        </div>
```

aduth (approved):

PR #298, opened 2022-02-18
Prepare 7.0 breaking changes

aduth:
This is a staging area for breaking changes we'd like to implement in a future major version release.

Currently, it includes:

>### Breaking Changes
>
>- Sass artifacts now require Dart Sass. The Ruby Sass and LibSass (C) compilers are no longer supported.
>- To improve performance of the default CSS bundle, responsive utility breakpoints other than `tablet:` have been disabled by default. Previously, `mobile-lg:` and `desktop:` utilities were also enabled. You can restore the original settings by overriding [the `$theme-utility-breakpoints` Sass variable](https://designsystem.digital.gov/documentation/settings/#utilities-settings).
>- Image URLs in stylesheets are no longer processed by `asset-path-if-exists`. For Rails integrations, consider using [the `$theme-image-path` Sass variable](https://designsystem.digital.gov/documentation/settings/#general-settings) in combination with [Sprockets `resolve_assets_in_css_urls` option](https://github.com/rails/sprockets-rails#initializer-options).
>- The deprecated `$image-path` Sass variable setting has been removed. Use `$theme-image-path` instead.
>- The deprecated `alerts/success-badge.svg` image has been removed. Use `alerts/success.svg` instead.


aduth:
It probably makes sense to release the next major version in tandem with USWDS v3, which is slated for release next month.

aduth:
> Other ideas:

I still like many of these ideas, but we can save them for a future major release.

aduth:
@zachmargolis I think I'd like to include these changes in the upcoming release. Could I tap you for a review?

CHANGELOG.md
```diff
@@ -55,6 +55,15 @@
 
 ## 6.3.2
```
zachmargolis, line 56: we want to move breaking into the next version so `7.0.0` here right? or are these pre-7 breaking changes?

zachmargolis (approved):
LGTM

CHANGELOG.md
aduth: Yeah, I don't know why I keep messing up these changelog notes across the pull requests 😖 I'm guessing a bad rebase conflict resolution.

aduth:
Visual regression failure is from the removal of the "Limiting Alert Width" documentation section, which is an expected breaking change.

PR #299, opened 2022-02-22
update radio image width

SammySteiner:
## Why
In order to have the correct sizing of svg icons (24px) for the illustrated ratio inputs, we need to update the width from 1 to 1.5 rem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Bug Fixes
+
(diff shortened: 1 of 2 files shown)


aduth (approved):
LGTM 👍

aduth:
Just to clarify, this is only meant to apply to mobile sizes, right? The tablet+ size is still set to 2rem width.

SammySteiner:
> Just to clarify, this is only meant to apply to mobile sizes, right? The tablet+ size is still set to 2rem width. (The lack of a visual regression failure caught my attention)

Yes, just mobile sizes.

PR #302, opened 2022-02-24
Remove redundant body background color

aduth:
**Why**: Because it's already set through USWDS.

See: https://github.com/uswds/uswds/blob/86636ea/src/stylesheets/settings/_settings-color.scss#L136

Before|After
---|---
Screen Shot 2022-02-24 at 9 09 39 AM|Screen Shot 2022-02-24 at 9 10 31 AM


zachmargolis (approved):
LGTM

PR #303, opened 2022-02-28
Reconcile focus styles with USWDS

aduth:
**Why**: To reduce the amount of generated styles for controlling focus appearance of fields, and to avoid error-prone reimplementation of focus styles by leveraging default styles and only overriding as needed.

Ideally there should be no visual changes associated with these revisions, which should be covered decently well with visual regression tests, as both buttons and radio/checkbox fields are documented with a full set of states.

### DIFF
--- a/src/scss/components/_buttons.scss
+++ b/src/scss/components/_buttons.scss
@@ -5,8 +5,7 @@ $button-stroke-big: inset 0 0 0 units($theme-button-stroke-width * 2);
 
   &:not([disabled]):focus,
   &:not([disabled]).usa-focus {
-    @include disable-default-focus-styles;
-    box-shadow: roundable-focus-outline-box-shadows();
+    outline-offset: units($theme-focus-offset);
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/scss/components/_buttons.scss
```diff
@@ -5,8 +5,7 @@ $button-stroke-big: inset 0 0 0 units($theme-button-stroke-width * 2);
 
   &:not([disabled]):focus,
   &:not([disabled]).usa-focus {
-    @include disable-default-focus-styles;
-    box-shadow: roundable-focus-outline-box-shadows();
+    outline-offset: units($theme-focus-offset);
```
aduth, line 8: We still need some overrides, since it appears that USWDS doesn't always use the configured theme variables.

See: https://github.com/uswds/uswds/issues/4543

aduth:
> Ideally there should be no visual changes associated with these revisions, which should be covered decently well with visual regression tests, as both buttons and radio/checkbox fields are documented with a full set of states.

🙏  visual regression tests, they caught a small issue with big outline buttons, which are styled with a thicker border in LGDS, but got lost in this refactor. I'll fix up shortly.

image

zachmargolis (approved):
LGTM

PR #305, opened 2022-03-21
Remove documentation styles from stylesheets

aduth:
**Why**: So that the design system artifact includes only styles which apply across projects.

In practice, this seems to only have the minor effect of altering the margin between documentation page heading and the subsequent paragraph (0.4rem to 0.5rem, or 0.5rem to 1rem), which does not seem worthwhile to maintain.

It's expected that visual regression tests would fail, and demonstrate the visual difference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Remove documentation-specific images from published package. ([#300](https://github.com/18F/identity-style-guide/pull/300))
 - Remove duplicate styles. ([#301](https://github.com/18F/identity-style-guide/pull/301))
 - Reconcile redundant focus styles with U.S. Web Design System. ([#302](https://github.com/18F/identity-style-guide/pull/302))
+- Remove documentation site styles from design system artifact. ([#305](https://github.com/18F/identity-style-guide/pull/305))
 
 ### Dependencies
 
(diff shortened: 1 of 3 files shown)


aduth:
> It's expected that visual regression tests would fail, and demonstrate the visual difference.

Example, showing reduction of margin between "Accordions" heading and subsequent paragraph from 0.5rem to 0.4rem.

image

aduth:
Another difference is that headings for content-only (no-sidebar) pages will no longer show as the extra-large "display" style. From what I can tell, this only impacts the ["Security Experience Principles" page](https://design.login.gov/security-experience-principles/).

Before|After
---|---
image|image

This is a bit more of a noticeable change, but I'm still inclined to not adjust to keep the styles, at least for the sake of consistency with other pages, which don't show a larger main heading. (FYSA @nickttng)

aduth:
I tried removing the classes from markup in [6793755](https://github.com/18F/identity-style-guide/pull/305/commits/6793755b3d7ee6ed4ed75450490783f2d2d480f1), but it turns out that these styles are actually inherited in-part from USWDS.

There are some issues with this and I think it would make sense to remove from there as well, so I created https://github.com/uswds/uswds/issues/4594 as a follow-up.

zachmargolis (approved):
LGTM

PR #307, opened 2022-03-25
Add SVG optimization lint to build

aduth:
See: https://github.com/18F/identity-style-guide/pull/306#discussion_r835409381

**Why:** So that images included with the design system are as lean as they can be, so that they load quickly.


zachmargolis (approved):
LGTM

aduth:
Visual regression test seems to be using the result of #306, likely because this branch was initially set to merge to that one. Going to assume it's all clear, since there's not intended to be visual changes here.

PR #311, opened 2022-04-08
Fix hidden search submit button text on small viewport

aduth:
Fixes regression of #259

**Why**: Because the submit button should have a visible text label.

Slack context: https://gsa-tts.slack.com/archives/C01710KMYUB/p1649431326074669

Live preview URL: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.app.cloud.gov/preview/18f/identity-style-guide/aduth-fix-hidden-search-text/components/search/

**Screenshots:**

Before|After
---|---
image|image

### DIFF
--- a/src/scss/components/_search.scss
+++ b/src/scss/components/_search.scss
@@ -6,8 +6,8 @@
   width: auto;
 }
 
-.usa-search .usa-search__submit-text {
-  position: static;
+.usa-search__submit-text {
+  display: block;
 }
 
 .usa-search--big [type='submit'],
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Unreleased
+## 6.4.1
 
 ### Bug Fixes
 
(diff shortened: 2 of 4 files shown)


src/scss/components/_search.scss
```diff
@@ -6,8 +6,8 @@
   width: auto;
 }
 
-.usa-search .usa-search__submit-text {
-  position: static;
```
aduth, line 10: The `position: static` was inherited from the original brochure site styles, but I can't see that it was actually doing anything. `static` is the default value, and it doesn't appear that this would be overriding anything.

nickttng (approved):

aduth:
Going to include the version bump in this branch so we can publish it right after merge.

aduth:
(Whoops, didn't mean to close 🙃 )

aduth:
Visual regression failure is expected, due to version number on the homepage:

image

PR #313, opened 2022-05-13
changelog: improvements, add style and markdown for checked and disab…

SammySteiner:
…led radios and checkboxes, LG-6230

## What
Adding styles and markdown content for displaying variations on radios and checkboxes that are both checked and disabled. 

## Why
Needed for LG-6230 when users see a form they previously submitted and can select more options but not unselect the one they previously selected. Those options are checked but they must also be disabled.

Example:
Screen Shot 2022-05-13 at 10 37 01 AM

### DIFF
--- a/src/scss/components/_inputs.scss
+++ b/src/scss/components/_inputs.scss
@@ -143,7 +143,7 @@ $input-select-margin-right: 1;
 
 .usa-radio__input--tile:disabled + .usa-radio__label--illustrated .usa-radio__image,
 .usa-checkbox__input--tile:disabled + .usa-checkbox__label--illustrated .usa-checkbox__image {
-  filter: invert(0%) sepia(1%) saturate(0%) hue-rotate(232deg) brightness(100%) contrast(100%);
+  filter: invert(45%) sepia(1%) saturate(0%) hue-rotate(232deg) brightness(100%) contrast(80%);
 }
 
 .usa-checkbox__input--tile + .usa-checkbox__label--illustrated,
--- a/src/scss/components/_inputs.scss
+++ b/src/scss/components/_inputs.scss
@@ -76,7 +76,7 @@ $input-select-margin-right: 1;
   border-radius: 50%;
 }
 
-.usa-radio__input:disabled+.usa-radio__label::before,
+.usa-radio__input:disabled + .usa-radio__label::before,
 .usa-checkbox__input:disabled + .usa-checkbox__label::before {
   box-shadow: inset 0 0 0 units($border-width) color('disabled');
 }
--- a/src/scss/components/_inputs.scss
+++ b/src/scss/components/_inputs.scss
@@ -84,6 +84,7 @@ $input-select-margin-right: 1;
 
 .usa-checkbox__input:checked + .usa-checkbox__label::before,
 .usa-checkbox__input:checked:disabled + .usa-checkbox__label::before {
+  background-color: color('disabled');
   background-size: units(1) auto;
   box-shadow: inset 0 0 0 units($border-width) color('disabled');
 }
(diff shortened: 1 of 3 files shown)


src/scss/components/_inputs.scss
```diff
@@ -70,6 +70,10 @@ $input-select-margin-right: 1;
 .usa-checkbox__input:disabled + .usa-checkbox__label {
   color: color('disabled');
   border-color: color('disabled');
+
+  .usa-checkbox__image {
+    filter: invert(45%) sepia(1%) saturate(0%) hue-rotate(232deg) brightness(100%) contrast(80%);
+  }
```
aduth, line 76: Are these styles necessary, since it appears they're the same as what would be applied below on lines 144-150?

src/scss/components/_inputs.scss
```diff
@@ -136,6 +141,14 @@ $input-select-margin-right: 1;
   }
 }
 
+.usa-radio__input--tile:disabled + .usa-radio__label--illustrated,
+.usa-checkbox__input--tile:disabled + .usa-checkbox__label--illustrated {
+  .usa-radio__image,
+  .usa-checkbox__image {
+    filter: invert(45%) sepia(1%) saturate(0%) hue-rotate(232deg) brightness(100%) contrast(80%);
+  }
+}
```
aduth, line 150: I might suggest avoiding the nesting here, so that (1) the image styles only apply for the correct combination of radio + radio or checkbox + checkbox, and (2) to reduce the size of the output ([see comparison](https://www.sassmeister.com/gist/82331cc1f75fa50d8f7f7938bfee1ad9)).

```suggestion
.usa-radio__input--tile:disabled + .usa-radio__label--illustrated .usa-radio__image,
.usa-checkbox__input--tile:disabled + .usa-checkbox__label--illustrated .usa-checkbox__image {
  filter: invert(45%) sepia(1%) saturate(0%) hue-rotate(232deg) brightness(100%) contrast(80%);
}
```

nickttng:
Is there a way that the checkmark could be more visible? The checkmark from the current example is barely visible (white foreground on light gray background, so, likely, the users won't be able to perceive what it represents.

In Mostyn's example, the checkbox has a darker background so that the white checkmark is more prominent.
Screen Shot 2022-05-13 at 10 46 05 AM

src/scss/components/_inputs.scss
SammySteiner: Not necessary, removing now. Thanks for catching.

src/scss/components/_inputs.scss
```diff
@@ -84,7 +84,9 @@ $input-select-margin-right: 1;
 
 .usa-checkbox__input:checked + .usa-checkbox__label::before,
 .usa-checkbox__input:checked:disabled + .usa-checkbox__label::before {
+  background-color: color('disabled');
   background-size: units(1) auto;
+  box-shadow: inset 0 0 0 units($border-width) color('disabled');
```
aduth, line 89: The background color and box shadow styling here should be limited to disabled fields. Currently, it's applying to all:

image

aduth, line 89: Also, these are currently only applying to checkboxes. Radio fields look significantly lighter in the current version of the branch. Is that intended?

image

src/scss/components/_inputs.scss
```diff
@@ -76,15 +76,19 @@ $input-select-margin-right: 1;
   border-radius: 50%;
 }
 
-.usa-radio__input:disabled + .usa-radio__label::before,
```
aduth, line 79: Should we want this to apply to radio buttons as well? As implemented, the border of the unselected, disabled radio is thicker than the unselected, enabled radio and lighter than the disabled checkbox.

image

image

image

Keeping this selector makes it look a bit more expected:

image

src/scss/components/_inputs.scss
```diff
@@ -76,15 +76,20 @@ $input-select-margin-right: 1;
...
+.usa-radio__input:disabled+.usa-radio__label::before,
```
aduth, line 79: I predict you'll get a lint error on this one.

```suggestion
.usa-radio__input:disabled + .usa-radio__label::before,
```

aduth (approved):
LGTM 👍

PR #317, opened 2022-06-03
Update devDependencies to latest versions

aduth:
**Why**: To avoid future maintenance burden, and to take advantage of latest features and bug/security fixes.

There's still some (non-concerning) security advisories after these upgrades, due to unpatched Gulp. It's unclear to me if Gulp is abandonware at this point, and may be worth considering moving away (already confusing how we have tasks split across Gulp, npm scripts, and Makefile 🤷 ).


zachmargolis (approved):
LGTM

PR #321, opened 2022-09-09
Upgrade USWDS to 3.x

aduth:
**Why**:

- To take advantage of the latest features and bug fixes
   - Notably, to be able to start taking advantage of new component-based cherry-picked imports
- To avoid additional maintenance burden due to version drift


src/scss/components/_navbar.scss
```diff
@@ -1,3 +1,6 @@
+@use 'uswds-core' as *;
+@use 'usa-nav';
```
zachmargolis, line 2: oooo namespaced imports in SCSS... would we ever want to update some of these to be `uswds-core.xyz` in the future?

aduth: We could. I think that's one of the main appeals of the namespaces, though I'm still finding it difficult to get used to it. I don't like the verboseness of prefixing everything, though I suppose the "everything is global" of older Sass has its own flaws.

src/scss/_uswds-theme.scss
```diff
@@ -0,0 +1,214 @@
...
+  'site-lighter-green': #a6d5af,
+  'site-lightest-green': #ebfcef,
+  // White
+  'site-white': #ffffff
+);
+
+@use 'uswds-core' with (
+  // ----------------------------------------
+  // Colors
+  // ----------------------------------------
```
aduth, line 61: I didn't like having to collapse all the settings here, since we have so many of them that we configure. I need to check syntax support, but I'd wonder if it'd be possible to at least do something like...

```scss
// src/scss/uswds-theme/_color.scss
$settings: (
  $theme-color-base-family: 'grey',
  // ...
);
```

```scss
// src/scss/uswds-theme/_index.scss
@use 'sass:map';
@use 'color';

@use 'uswds-core' with map.merge(color.$settings /* , ... */ );
```

aduth: I tested this and it did not work. While the `with` parameter looks like it's provided as a map value, that value must be given using the map syntax, and not as a map variable / result of a map operation like `map.merge`.

aduth:
Hm, looks like my celebration of "All checks have passed" may be premature, since it looks like the last commit didn't run in CircleCI. I'll try to see what happened.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ## 6.7.0
 
+### Breaking Changes
```
zachmargolis, line 3: is the 6.7.0 release pending still? or do we need to cut a new release?

zachmargolis (approved):
LGTM

nickttng (approved):

PR #323, opened 2022-09-28
Add redirect for /components/ path

aduth:
**Why**: Since its sensible that the top-level path is available at its index base, and to resolve a 404 from [the IdP's frontend documentation](https://github.com/18F/identity-idp/blob/main/docs/frontend.md) ([related discussion](https://github.com/18F/identity-idp/pull/7037#discussion_r982459507)).

### DIFF
--- a/docs/_components/index.md
+++ b/docs/_components/index.md
@@ -1,0 +1,4 @@
+---
+permalink: /components/
+redirect_to: /components/accordions/
+---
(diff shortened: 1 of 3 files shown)


aduth:
Thank goodness for visual regression specs, this caught that the new `index.md` would cause an entry to incorrectly be added to the navigation for components.

Screen Shot 2022-09-28 at 11 04 50 AM

I can probably move it elsewhere like `docs/components.md`.

aduth:
Closing since we're in the process of spinning down the site.

For the IdP 404, see: https://github.com/18F/identity-idp/pull/8159

PR #324, opened 2022-10-05
Migrate Federalist to Cloud.gov Pages

aduth:
**Why**: [Federalist is becoming Cloud.gov Pages](https://cloud.gov/pages/federalist-migration/), and as such we'll need to update configurations accordingly.

The webhook settings have already been updated, so these changes are updating documentation and code references to Federalist.


zachmargolis (approved):
LGTM

PR #325, opened 2022-12-12
Update Text Input styling to adhere to design specification

aduth:
In implementing LG-8066, I noticed that text inputs and hints were not spaced correctly per the Figma Design System reference. The changes here eliminate all margins around the hint, and reduces the top margin of text fields from 0.5rem to 0.25rem.

Variant|Before|After
---|---|---
Default|image|image
With Hint|image|image

Live preview: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.sites.pages.cloud.gov/preview/18f/identity-style-guide/aduth-hint-spec/components/form-fields/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Improvements
+
+- Update Text Input styling to adhere to design specification.
+
 ## 6.5.0
 
 ### Improvements
--- a/src/scss/elements/form-controls/_all.scss
+++ b/src/scss/elements/form-controls/_all.scss
@@ -1,1 +1,1 @@
-@import 'global';
+@import 'text-input';
--- a/src/scss/elements/form-controls/_all.scss
+++ b/src/scss/elements/form-controls/_all.scss
@@ -1,1 +1,1 @@
-@import 'text-input';
+@import 'global';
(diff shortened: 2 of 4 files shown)


aduth:
Visual regression failures are expecting since we're adjusting vertical spacing of form fields.

Example:

image

nickttng (approved):
LGTM. Do we want to apply the same to the [dropdowns](https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.sites.pages.cloud.gov/preview/18f/identity-style-guide/aduth-hint-spec/components/form-fields/#dropdowns)?

nickttng:
Got it. If it's a light lift to update across the board, I think it'd be good to have spacing consistency between any label and input field.

aduth:
> [the `%block-input-styles` Sass placeholder](https://github.com/uswds/uswds/blob/360865114a749c3f6859951947f1f21ef67aaf32/packages/uswds-core/src/styles/placeholders/_forms.scss#L17-L28), which is (a) not directly customizable

Actually, I take that back. Apparently in Sass, if you define a placeholder which already exists, it doesn't necessarily "override" what existed previously, but it does output an additional set of styles which would effectively take precedence over the initial definition, which should give us the effect we want.

Example: https://www.sassmeister.com/gist/d12223c9afc58b44c411146e54053df4

aduth:
Ok, the changes in 2871d86 should get us what we want. The updated preview should be available shortly.

aduth:
I'm going to go ahead and merge this since I want to start implementing "Big" inputs, but let me know if you spot any issues @nickttng and I'll be happy to follow-up on them.

nickttng:
Thank you, Andrew. I'll keep an eye out for the "Big" inputs.

PR #326, opened 2022-12-21
Add "Big" variant of Text Input, Textarea

aduth:
Adds new "Big" variants of Text Input and Textarea components to reflect Figma LGDS reference, and to support migration toward design system inputs in the IdP.

Live preview: https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.sites.pages.cloud.gov/preview/18f/identity-style-guide/aduth-big-text-inputs/components/form-fields/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ### Improvements
 
-- Update Text Input styling to adhere to design specification.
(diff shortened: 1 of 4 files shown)


nickttng (approved):
Both look good to me. Any particular feedback that you're looking for?

aduth:
> Both look good to me. Any particular feedback that you're looking for?

Nothing in particular! Unless you notice anything off between the specification and the implemented field.

aduth:
Visual regression test failure is expected for the new documented examples:

image

PR #332, opened 2023-03-10
Replace Node mkdirp package with native option

aduth:
Why: Fewer dependencies

Docs: https://nodejs.org/dist/latest-v12.x/docs/api/fs.html#fs_fs_mkdir_path_options_callback


zachmargolis (approved):
LGTM

PR #333, opened 2023-03-21
Show 404 page for documentation site

aduth:
As part of the current effort to deprecate design.login.gov, show a 404 page in place of site contents, so that a user has resources to learn more about Login.gov or receive support.

**Live preview:** https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.sites.pages.cloud.gov/preview/18f/identity-style-guide/aduth-docs-404/404.html

This is derived from the Login.gov main site 404 page: https://www.login.gov/404.html

Next steps:

- ~Revisions to `_config.yml` will not be committed, they are for demonstration only in the preview environment. The plan would be to configure these through Cloud.gov Pages interface~
   - Configuration in 6982fbd8426b31686d4fb435e03ed8604e0de047 will be applied in Cloud.gov Pages site admin
- Excluding content in the live environment will allow us to continue using the documentation site for internal testing and pull request previews
   - Follow-on work should plan to remove most documentation aside from component previews
   - At some point, we may remove all site content altogether, or simplify it such that we can remove the Jekyll build pipeline, or move everything to [Login.gov Component Previews environment](https://idp.dev.identitysandbox.gov/components/), or move package contents to be a [workspace package within the `18f/identity-idp` repository](https://github.com/18F/identity-idp/tree/main/app/javascript/packages).

~Draft while I sort out why Cloud.gov Pages doesn't show the 404 page by default.~ See https://github.com/18F/identity-style-guide/pull/333#issuecomment-1479572932


src/img/login-gov-logo-rev.svg
aduth, line 1: @nickttng Curious your thoughts on this in-place replacement of the 'reverse' version of the logo. I don't think we use / want the all-white logo?

aduth:
Maybe we need to normalize the intrinsic sizes of those two images.

aduth:
> Maybe we need to normalize the intrinsic sizes of those two images.

In 608fde9, I recreated the reverse logo by taking the original and swapping out all the `#112E51` with `#ffffff` 🤷

docs/404.md
```diff
@@ -0,0 +1,3 @@
+---
+layout: 404
```
zachmargolis, line 2: does the `docs` dir already go to top level? do we need to add a permalink so that it's in the right place to be a 404 page for CG pages?
```suggestion
layout: 404
permalink: /404.html
```

zachmargolis (approved):
LGTM

docs/404.md
aduth: I think it's fine without? It seems to be [available at `/404.html`](https://federalist-2f194a10-945e-4413-be01-46ca6dae5358.sites.pages.cloud.gov/preview/18f/identity-style-guide/aduth-docs-404/404.html) already without specifying `permalink`.

PR #334, opened 2023-04-03
Standardize on "tablet" layout breakpoint

aduth:
Fixes issue where documentation pages with sidebar appear incorrectly stacked. Consistent with changes proposed in #298 (specifically 24ed5612793536cd50d1c2e3090ff47cc8cd2b2d), standardize on the "tablet" breakpoint.

This breakage of the documentation site wasn't caught by visual regression specs since the screen size for visual regression specs was set to a size which wouldn't have been affected by this change.

### DIFF
--- a/docs/_layouts/main.html
+++ b/docs/_layouts/main.html
@@ -8,11 +8,11 @@ layout: base
 
 {% assign sidenav = site.data.nav[page.sidenav] | default: page.sidenav %}
 
-<main class="usa-layout-docs usa-section" id="main-content">
+<main class="usa-section" id="main-content">
   <div class="grid-container">
     {% if sidenav %}
       <div class="grid-row grid-gap">
-        <aside class="usa-layout-docs__sidenav desktop:grid-col-3">
+        <aside class="display-none tablet:display-block tablet:grid-col-3">
           <nav>
             <ul class="usa-sidenav">
               {%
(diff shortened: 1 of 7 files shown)


docs/_layouts/main.html
```diff
@@ -8,11 +8,11 @@
 
 {% assign sidenav = site.data.nav[page.sidenav] | default: page.sidenav %}
 
-<main class="usa-layout-docs usa-section" id="main-content">
+<main class="usa-section" id="main-content">
   <div class="grid-container">
     {% if sidenav %}
       <div class="grid-row grid-gap">
-        <aside class="usa-layout-docs__sidenav desktop:grid-col-3">
+        <aside class="display-none tablet:display-block tablet:grid-col-3">
```
aduth, line 15: These changes move away from the internal-ish [`usa-layout-docs` styles](https://github.com/uswds/uswds/blob/develop/packages/usa-layout-docs/src/styles/_usa-layout-docs.scss), which should have an accessibility benefit by avoiding [issues concerning `order` styles](https://developer.mozilla.org/en-US/docs/Web/CSS/order#accessibility_concerns). Instead, the markup is duplicated so that it still appears below content as expected at small screens, but reflecting that in the markup as well.

zachmargolis: What's our timeline for spinning down this site? Should we just start removing content from it so we don't have to maintain it?

zachmargolis (approved):
LGTM

docs/_layouts/main.html
aduth: > What's our timeline for spinning down this site? Should we just start removing content from it so we don't have to maintain it?

Still nailing down specifics on this. #333 will prep the production site to 404. Once that's in place, I was going to remove many of the non-component content pages, but may want to continue keeping around some of the rest of the documentation for internal development reference, pull request preview links, and visual regression specs. So it'd continue to be useful to have this working, at least for the short term.

PR #337, opened 2023-04-14
Remove deprecated hint class, update Memorable Date guidance

aduth:
This removes the `.usa-form-hint` class, which was deprecated in USWDS v2 ([documentation](https://designsystem.digital.gov/documentation/migration-v2/#inputs)). See https://github.com/18F/identity-dashboard/pull/634 for removal of the only downstream usage I could find.

The class was also used internally in the Memorable Date component guidance. I took the opportunity to update the markup to align to the new select-based dropdown. This documentation may be short-lived anyways, what with https://design.design.gov slated for deprecation.

Before|After
---|---
image|image


nickttng (approved):

zachmargolis (approved):
LGTM

PR #341, opened 2023-04-18
Add README section for configuring theme variables

aduth:
Since it's worth reaffirming that the design system already preconfigures settings, as well as anticipate that it may be necessary to configure in some cases ([example](https://github.com/18F/identity-idp/pull/8093/files#diff-2ffaf854e681b9e2cdfb8af6fb64a607d0580b978900873f2988d45d1161fcfa)).


zachmargolis (approved):
LGTM

PR #339, opened 2023-04-18
Bump dependencies to latest

aduth:
As a maintenance task in preparation for upcoming release.

A couple dependencies were removed as they're no longer used or necessary:

- `gulp-replace`
- `eslint-config-airbnb-base`

_Before:_

```
$ npm i

up to date, audited 1469 packages in 1s

9 vulnerabilities (1 moderate, 8 high)
```

_After:_

```
$ npm i

up to date, audited 1423 packages in 1s

6 high severity vulnerabilities
```

PR #342, opened 2023-04-18
Limit size of published Sass content

aduth:
Why? So that the package isn't exceedingly large or include files that are not necessary.

Previously, we would copy Sass files from `src/assets/scss` to `dist/assets/scss`, but with the updated usage recommending to use the `packages/` entrypoint (see README.md), the `dist/assets/scss` directory is not necessary.

Additionally, when running `npm publish --dry-run`, I was seeing that there was some `node_modules` files being included, which turns out to be an upstream issue for USWDS (see https://github.com/uswds/uswds/pull/5245). I've excluded them here as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ build: build-docs build-assets build-package
 build-docs:
 	JEKYLL_ENV=production bundle exec jekyll build
 
-build-assets: build-sass-and-js build-fonts build-images copy-scss build-sass-packages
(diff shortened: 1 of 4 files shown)


zachmargolis (approved):
LGTM

aduth:
Based on the build failure, I guess I was wrong with my assumption at https://github.com/18F/identity-style-guide/pull/339#issuecomment-1513335517 that the issue would work itself out after the initial merge 😬

PR #343, opened 2023-04-19
Fix screenshot regression tests

aduth:
Fixes an issue where visual regression tests are failing due to an issue introduced with the upgrade of dependencies, including Puppeteer. See https://github.com/18F/identity-style-guide/pull/339#issuecomment-1513335517 for more context, where the issue continued into subsequent pull requests (#342, for example).

This was just a guess at a solution to try to wait until the page load was more "stable" before taking the screenshot, but it appears to be working on first pass.

Related documentation: https://pptr.dev/api/puppeteer.page.goto/#remarks

### DIFF
--- a/test/screenshot.test.js
+++ b/test/screenshot.test.js
@@ -38,7 +38,7 @@ async function toggleHiddenContent() {
 }
 
 async function getScreenshot(url) {
-  await page.goto(url);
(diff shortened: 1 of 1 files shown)


zachmargolis (approved):
LGTM

PR #344, opened 2023-04-20
Move global elements styles to elements package

aduth:
**Why?**

- It's an appropriate location for these styles, since [the upstream package](https://github.com/uswds/uswds/tree/develop/packages/uswds-elements/src/styles) contains these sorts of global HTML element styles
- It makes it easier to override the `uswds` package as [documented in the README](https://github.com/18F/identity-style-guide#optimization), since otherwise the style needs to be repeated wherever individual components are being forwarded ([see IdP integration example](https://github.com/18F/identity-idp/commit/c53dff89f93752b280ff00e3d105d7707a9a64bf))

### DIFF
--- a/src/scss/packages/_uswds-elements.scss
+++ b/src/scss/packages/_uswds-elements.scss
@@ -1,0 +1,7 @@
+@use 'uswds-core' as *;
+@forward './uswds-elements/index';
+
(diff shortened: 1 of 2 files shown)


zachmargolis (approved):
LGTM

PR #345, opened 2023-04-24
Fix alert padding and border radius (USWDS 3 upgrade regression)

aduth:
Why?

- USWDS upgrade regressed two intended styles:
  - Overriding horizontal padding
  - Setting border radius on the alert

Before|After
---|---
image|image

### DIFF
--- a/src/scss/packages/_usa-alert.scss
+++ b/src/scss/packages/_usa-alert.scss
@@ -3,20 +3,20 @@
 
 $alerts: 'success', 'warning', 'error', 'info', 'emergency';
 
-.usa-alert__body {
-  @include u-padding-x($theme-alert-padding-x); // See: https://github.com/uswds/uswds/issues/5252}
-}
-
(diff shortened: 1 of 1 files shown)


src/scss/packages/_usa-alert.scss
```diff
@@ -4,12 +4,19 @@
 $alerts: 'success', 'warning', 'error', 'info', 'emergency';
 
 .usa-alert {
-  border-radius: radius('lg');
+  &,
+  & .usa-alert__body {
+    border-radius: radius('lg');
```
aduth, line 9: It appears that as of USWDS 3, the alert background is applied to both `.usa-alert` as well as `.usa-alert__body`, so adding the `border-radius` to `.usa-alert` would cause the appearance of squared corners, since `.usa-alert__body` would previously have not had the same border radius.

zachmargolis (approved):
LGTM

PR #351, opened 2023-04-28
Rename "login-gov-logo-rev" to "login-gov-logo-reverse"

aduth:
**Why?**

1. Clearer name and more consistent with existing design system terminology ([examples](https://github.com/uswds/uswds/blob/29eebc4/packages/uswds-core/src/styles/settings/_settings-color.scss#L138-L147))
2. Technically someone might consider the changed logo in #333 to be a breaking change since it's visually different
3. Get all our breaking changes in for 7.0.0


nickttng (approved):
em ot doog skooL 😄

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - The deprecated `.usa-form-hint` CSS class has been removed. Use `.usa-hint` instead.
 - Removed `usa-alert__paragraph` alert helper class. Use [`measure-3` measure utility](https://designsystem.digital.gov/utilities/paragraph-styles/#max-width) instead.
 - The Spinner component has been removed (`.lg-spinner`). A new "Loading" component will be made available in a future release.
+- The all-white Login.gov logo `login-gov-logo-rev.svg` has been replaced with a new color-reversed version `src/img/login-gov-logo-reverse.svg`.
```
zachmargolis, line 20: the image file in this PR says its only renamed, not acutally changed, is that because we "snuck" that in with https://github.com/18F/identity-style-guide/pull/333 ?

zachmargolis (approved):

CHANGELOG.md
aduth: Yeah, the changelog here reflects both the revisions to the image itself in #333 as well as the rename. The changes in #333 probably should have been noted in the CHANGELOG, though it's debatable as being "breaking" since it's more a correction that shouldn't require any other downstream changes.

PR #352, opened 2023-05-01
Fix header navigation current link styling regression

aduth:
This resolves a regression introduced as part of the changes in the 7.0.0 release where the "current" link would be shown with a blue underline instead of a red underline.

Example from https://design.login.gov/brand/ :

image

Compare to https://login.gov/what-is-login/ :

image

After discussing with @nickttng , the changes here are a bit more of an extensive "reset" to USWDS stock styling, with minimal overrides to apply expected customizations:

- Remove bold text styling from links other than the current header navigation linik
- Change bar underline from blue ("primary" token) to red ("secondary" token)
- Reduce horizontal padding between header navigation items from 2rem (32px) to 1.5rem (24px)
- Reduce margins of logo
- Increase vertical padding of side navigation and mobile navigation items from 0.5rem (8px) to 1rem (16px)
- Change hover background color for side navigation and mobile navigation items from `base-lightest` (`#f0f0f0`) to `primary-lightest` (`#f2f9ff`)
- Remove borders between submenu links in side navigation and mobile navigation
- **Edit:** More revisions added based on feedback:
   - Remove border-radius from current link bar in side navigation and mobile navigation
   - Reduce size of mobile navigation "X" button
- **Edit 2:** More revisions based on further testing with downstream projects:
   - Remove border below header when followed by dark section (e.g. hero)
   - Increase font size of header navigation to from 13px to 14px

The idea of resetting to the USWDS is...

- Better understand divergence with USWDS
- Eliminate divergence where possible to reduce maintenance overhead and CSS size
- Align divergence with specific CSS selectors to reduce risk of future breakage

Open questions:

- Possible other revisions could include...
   - ~Changing side navigation and mobile navigation line to extend the full height of the list item~ **Edit:** Based on https://github.com/18F/identity-design-system/pull/352#pullrequestreview-1408054717, keeping height as-is, but removed border radius in 9f59575.
   - ~Reducing size of mobile navigation "X" button~ **Edit:** Added in e3a5d1f

To review:

- Header logo
- Header navigation
- Side navigation (impacted because it shares many of its stylings with mobile header navigation)

Component|Before|After
---|---|---
Header|image|image
Side navigation|image|image
Mobile navigation|design login gov_components_side-navigation_(iPhone XR)|localhost_4000_components_side-navigation_(iPhone XR)

### DIFF
--- a/src/scss/packages/_usa-sidenav.scss
+++ b/src/scss/packages/_usa-sidenav.scss
@@ -15,6 +15,7 @@
 
     &::after {
       background-color: color('secondary');
+      border-radius: 0;
     }
 
     &:hover {
--- a/src/scss/packages/_usa-sidenav.scss
+++ b/src/scss/packages/_usa-sidenav.scss
@@ -21,6 +21,10 @@
       color: color('primary-dark');
     }
   }
+
+  > .usa-sidenav__item > .usa-sidenav__sublist {
+    border-top: units(1px) solid color('base-lighter');
+  }
 }
 
 .usa-sidenav__sublist a:not(.usa-button) {
--- a/src/scss/packages/_usa-sidenav.scss
+++ b/src/scss/packages/_usa-sidenav.scss
@@ -7,28 +7,20 @@
 
     &:hover {
       background-color: color('primary-lightest');
-      color: unset;
     }
   }
 
   .usa-current {
-    color: $theme-color-primary-dark;
+    color: color('primary-dark');
 
-    @include add-bar($theme-sidenav-current-border-width, 'secondary', 'left', 0, 0, 0);
-
-    @include at-media('tablet') {
-      @include add-bar($theme-sidenav-current-border-width, 'secondary', 'left', 0, 0, 0);
+    &::after {
+      background-color: color('secondary');
     }
 
     &:hover {
-      color: $theme-color-primary-dark;
+      color: color('primary-dark');
     }
   }
-}
-
-.usa-sidenav--sticky {
-  position: sticky;
-  top: units(2);
 }
 
 .usa-sidenav__sublist a:not(.usa-button) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@
 
 ### Bug Fixes
 
-- Fix header navigation current link styling regression in 7.0.0 release.
+- Fix header navigation current link styling regression in 7.0.0 release. ([#352](https://github.com/18F/identity-design-system/pull/352))
+
+### Breaking Changes for Undocumented APIs
+
+_Note:_ While these are backwards-incompatible changes, the major version is not being changed because they impact features which were never part of a publicly-documented API.
+
+- Remove `.usa-sidenav--sticky` CSS class. ([#352](https://github.com/18F/identity-design-system/pull/352))
 
 ## 7.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Bug Fixes
 
 - Fix header navigation current link styling regression in 7.0.0 release. ([#352](https://github.com/18F/identity-design-system/pull/352))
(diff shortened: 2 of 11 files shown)


src/scss/packages/_usa-sidenav.scss
```diff
@@ -7,30 +7,22 @@
...
     }
 
     &:hover {
-      color: $theme-color-primary-dark;
+      color: color('primary-dark');
     }
   }
 }
 
-.usa-sidenav--sticky {
```
zachmargolis, line 29: was this unused or part of USWDS now?

zachmargolis (approved):
LGTM

src/scss/packages/_usa-sidenav.scss
aduth: I'm glad you caught it, I had an internal debate about this, since it is unused across our projects, and doesn't exist in USWDS, but it's still technically a breaking change. I'm feeling a bit more comfortable with ripping out unused legacy cruft without bumping the major version, particularly for things which aren't documented anywhere. But I'm open to dissenting opinions!

zachmargolis: while there *could* be other consumers of this project besides us, AFAIK there aren't any. I think it's fine to break something like this, but it might be worth a release notes item

nickttng (approved):
Overall, it looks solid to me. 

> Possible other revisions could include...
> - Changing side navigation and mobile navigation line to extend the full height of the list item

I am okay not having to extend if it means less maintenance cost. If anything to be consistent, I think updating the `border-radius` to `0` would align better with the desktop nav style

> - Reducing size of mobile navigation "X" button

Okay with reducing the resize of the "X" button as long it doesn't compromise the tap area size.

src/scss/packages/_usa-sidenav.scss
aduth: I'll just restore it and then add it to a running set of breaking changes to be included in a future major release, like #298.

aduth: I changed my mind. Let's keep the removal. I added a "Breaking Changes for Undocumented APIs" section to the CHANGELOG in 151c794, with a note about why it's not bumping the major version.

aduth:
I removed the pill border radius and reduced the mobile navigation close icon in 9f59575 and e3a5d1f respectively.

Screenshots showing before the change and after. I'll update the original pull request comment with Before/After for "live" vs. proposed changes.

Before|After
---|---
before|localhost_4000_components_side-navigation_(iPhone XR) (1)
before|image

aduth:
I'm going to test this on local copies of the brochure site, developer site, and IdP to check how well some of changes are working out. I've pushed a few commits already based on findings, will put back to draft temporarily while I work through any other findings.

aduth:
Ok, I've finished testing on both desktop and mobile across the brochure site, help center, partner site, developer site, and IdP, and feeling pretty good now after the latest set of changes. It required a bit of extra work to account for things such as partner site header arrangement with search and mobile collapsible menu items. Overall it'll help simplify a lot of the markup and custom styles in those sites as well, which is nice.

The [preview site](https://federalist-340d8801-aa16-4df5-ad22-b1e3e731098e.sites.pages.cloud.gov/preview/18f/identity-design-system/aduth-fix-nav/) includes the latest changes. If it's useful, I could also publish a beta release to get a preview going for those other sites, though maybe it's just as well to go ahead and publish the patch release and we can follow-up with another patch release if there's any further corrections to make.

aduth:
I'm going to plan to merge this as-is and publish a beta release, which will allow us to test it more exhaustively in preview branches for the brochure site, partner site, and developer site.

aduth:
> I'm going to plan to merge this as-is and publish a beta release, which will allow us to test it more exhaustively in preview branches for the brochure site, partner site, and developer site.

See:

- https://github.com/18F/identity-site/pull/1073
- https://github.com/18F/identity-dev-docs/pull/331
- https://github.com/18F/identity-idp/pull/8093

PR #353, opened 2023-05-01
Run regression spec against preview site

aduth:
**Why?** So that it'll continue to work after the main design.login.gov site is decommissioned.

Also updates a few other internal references to https://design.login.gov which would break after the site is taken offline.

### DIFF
--- a/test/screenshot.test.js
+++ b/test/screenshot.test.js
@@ -10,7 +10,8 @@ const { getURLsFromSitemap } = require('./support/sitemap');
 const { writeFile, mkdir } = fsPromises;
 
 const LOCAL_HOST = `http://localhost:${process.env.JEST_PORT}`;
-const REMOTE_HOST = 'https://design.login.gov';
+const REMOTE_HOST =
+  'https://federalist-340d8801-aa16-4df5-ad22-b1e3e731098e.sites.pages.cloud.gov/site/18f/identity-design-system';
 const DIFF_DIRECTORY = join(__dirname, '../tmp/results/screenshot-diff');
 
 async function stubAnimations() {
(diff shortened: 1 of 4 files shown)


aduth:
There'll be an initial visual regression failure here since the "live" site doesn't yet have the change from ad24970 and will include the site's `baseurl` in code snippets (see example [before](https://federalist-340d8801-aa16-4df5-ad22-b1e3e731098e.sites.pages.cloud.gov/site/18f/identity-design-system/components/badges/) and [after](https://federalist-340d8801-aa16-4df5-ad22-b1e3e731098e.sites.pages.cloud.gov/preview/18f/identity-design-system/aduth-regression-preview-site/components/badges/)).

test/screenshot.test.js
```diff
@@ -10,7 +10,8 @@ const { getURLsFromSitemap } = require('./support/sitemap');
 const { writeFile, mkdir } = fsPromises;
 
 const LOCAL_HOST = `http://localhost:${process.env.JEST_PORT}`;
-const REMOTE_HOST = 'https://design.login.gov';
+const REMOTE_HOST =
+  'https://federalist-340d8801-aa16-4df5-ad22-b1e3e731098e.sites.pages.cloud.gov/site/18f/identity-design-system';
```
zachmargolis, line 14: WDYT of trying to have this do a `git checkout main` or something instead of continuing to rely on federalist? My expectation is that by shutting down `design.login.gov` we'd remove everything published?

zachmargolis (approved):
LGTM

test/screenshot.test.js
aduth: For the near term, I'm expecting we'll continue to use Federalist to be able to use the branch preview functionality, even if the main https://design.login.gov site will start to show a 404 (which, per https://github.com/18F/identity-design-system/pull/333#issuecomment-1528093907, is still going to be hosted by Federalist).

zachmargolis: sounds good!

PR #355, opened 2023-05-09
Remove uswds-packages symlink

aduth:
**Why?**

- Simpler
- We don't need it, we can point directly to node_modules for the only place we use it now (Sass loadPaths)
- May have better cross-OS compatibility (i.e. Windows)


zachmargolis (approved):
LGTM but what's windows?

PR #356, opened 2023-05-09
Override USWDS Sass packages at source

aduth:
This builds upon #336 to further restructure the folder structure to align to how files are managed in the upstream project.

**Why?**

- Improved performance. Improves build time for the default package by nearly 50% (8.8s to 4.7s)
   - See related comment at https://github.com/uswds/uswds/pull/5241#issuecomment-1540686386
- Better maintainability. Avoids needing to keep `pacakges/_uswds.scss` in sync as the upstream USWDS adds new components over time.
- Coverage for different import semantics. Specifically, supports both import of the named package, as well as [the "Using Package Source" performance optimization](https://designsystem.digital.gov/documentation/migration/#using-package-source).
- Clearer separation. Overrides are strictly separate from the base styles.

Unfortunately the diff didn't turn out as clean as I'd hoped, despite the fact that 99% of changes are file renames.

Essentially, this converts:

**Before:**

```
src/
 - scss/
  - packages/
   - _usa-accordion.scss
```

```scss
// _usa-accordion.scss

@forward './usa-accordion/index';

.usa-accordion {
  // ...customizations...
}
```

**After:**

```
src/
 - scss/
  - packages/
   - _usa-accordion/
    - src/
     -  _styles.scss
     - _overrides.scss
   - _usa-accordion.scss
```

```scss
// _usa-accordion.scss

@forward 'usa-accordion/index';
@forward 'usa-accordion/src/overrides';
```

```scss
// usa-accordion/src/_styles.scss

@forward 'usa-accordion/src/styles/index';
@forward './overrides';
```

```scss
// usa-accordion/src/_overrides.scss

.usa-accordion {
  // ...customizations...
}
```

The same general principle of intercepting USWDS files applies, but now applies for both the top-level import (`@forward 'usa-accordion'`) as well as the package source (`@forward 'usa-accordion/src/styles'`).

The only meaningful code change here was to eliminate `$border-width` variable in some of the SCSS files. I was running into an error message that seemed to be caused by multiple forwarded modules each defining the variable. Unclear why this wasn't an issue previously 😬 

```
Error: Two forwarded modules both define a variable named $border-width.
  ╷
1 │ @forward 'usa-checkbox/index';
  │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ original @forward
2 │ @forward 'usa-checkbox/src/overrides';
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ new @forward
  ╵
[...]

PR #357, opened 2023-05-26
Fix Process List "Big" variant showing connected line

aduth:
~Merges to (blocked by) #356~

Fixes an issue where Process List "Big" variant would show with the connected line when not rendered in prose content.

Issue can be seen here ("Big" should not show line): https://idp.dev.identitysandbox.gov/components/inspect/process_list/preview

Issue can't be seen here, since design system documentation is shown in `.usa-prose`: https://design.login.gov/components/process-list/#big

The underlying issue is a CSS specificity conflict, where the styles responsible for [unsetting the border width](https://github.com/18F/identity-design-system/blob/76d17e736e7c446476e22246ce1230bbd37745db/src/scss/packages/_usa-process-list.scss#L11-L13) were only strong enough to override [the previous border width styling](https://github.com/18F/identity-design-system/blob/76d17e736e7c446476e22246ce1230bbd37745db/src/scss/packages/_usa-process-list.scss#L55) when rendered inside `.usa-prose`, but not when rendered outside `.usa-prose`. The changes here make it so that the "Big" border is only applied to the connected process list.

**Testing Instructions:**

1. Go to https://federalist-340d8801-aa16-4df5-ad22-b1e3e731098e.sites.pages.cloud.gov/preview/18f/identity-design-system/aduth-fix-process-list-big-connnected/process-list/#big
2. Observe the Big Process list shows without border
3. In DevTools console, enter `document.querySelector('.usa-prose').classList.remove('usa-prose')`
4. Observe the Big Process list still shows without border

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 - Fix appearance of borders in segmented "Big" button groups. ([#359](https://github.com/18F/identity-design-system/pull/359))
 - Improve compilation time for default package by upwards of 50%. ([#356](https://github.com/18F/identity-design-system/pull/356))
 
+### Bug Fixes
+
+- Fix issue where the Process List "Big" variant would show with the connected line when not rendered in prose content.
+
 ## 7.0.1
 
 ### Bug Fixes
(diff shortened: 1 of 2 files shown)


eric-gade:
@aduth I think the federalist link for testing the fix per the instructions expired

eric-gade (approved):
code changes look good to me

aduth:
> @aduth I think the federalist link for testing the fix per the instructions expired

Thanks for catching! It got outdated in some of the path shuffling with #358. I updated the link in the original comment.

eric-gade:
Ok sweet, and the live test is working on my end. All looks good.

zachmargolis (approved):
LGTM

PR #358, opened 2023-06-02
Remove most content from documentation output

aduth:
This pull request removes most of the content from the Jekyll site output from this repository. Now that https://design.login.gov is no longer published, the value provided by the content is now largely as a preview for components in Federalist branch previews, and for visual regression testing of those components. (This, too, may change over time)

Live preview: https://federalist-340d8801-aa16-4df5-ad22-b1e3e731098e.sites.pages.cloud.gov/preview/18f/identity-design-system/aduth-remove-docs/


zachmargolis (approved):
LGTM

PR #359, opened 2023-06-08
Improve styling for segmented "big" button group

aduth:
~Merges to (blocked by) #356~

**Why?** LGDS styles default button stroke width as 1px, and has custom styles to override stroke width to 2px for big buttons. USWDS button group styles use the stroke width, but are not aware of our customizations to double the width for big buttons, so we need to manually override these styles.

In practice, this makes a small difference on the appearance of a borders buttons.

Before|After
---|---
before|after

Live preview URL: https://federalist-340d8801-aa16-4df5-ad22-b1e3e731098e.sites.pages.cloud.gov/preview/18f/identity-design-system/aduth-button-groups-border/components/button-groups/

Related IdP usage: https://github.com/18F/identity-idp/pull/8542

LGDS big button customizations:

https://github.com/18F/identity-design-system/blob/1a488efdb6ff4ca1f0a05873440d60be93ec8c5e/src/scss/packages/_usa-button.scss#L4

Relevant USWDS styles: https://github.com/uswds/uswds/blob/8736f9305ee7547b7828fc1d625d83f458c6b44c/packages/usa-button-group/src/styles/_usa-button-group.scss#L80-L99

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Improvements
 
+- Fix appearance of borders in segmented "Big" button groups. ([#359](https://github.com/18F/identity-design-system/pull/359))
 - Improve compilation time for default package by upwards of 50%. ([#356](https://github.com/18F/identity-design-system/pull/356))
 
 ## 7.0.1
--- a/docs/_components/button-groups.md
+++ b/docs/_components/button-groups.md
@@ -1,0 +1,41 @@
+---
+title: Button Groups
+---
+
+{% include helpers/base-component.html component="button-group" %}
+
+## Segmented
+
+### Default
+
+{% capture example %}
+<ul class="usa-button-group usa-button-group--segmented">
+  <li class="usa-button-group__item">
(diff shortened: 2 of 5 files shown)


nickttng:
@aduth - I have an item to run by you

In the screenshots below, the "button extra" button is misaligned with others because of the length of its button content, extending the vertical height.

We could provide a caveat or text guidance to it, mitigating this kind of misalignment scenario. Or, If it is possible, for buttons to be vertically aligned, regardless of text length. Open to ideas. 

|When on mobile (less than 440px width)|When zoomed more than 400% on Chrome|
|---|---|
|IMG_0022|Screen Shot 2023-06-08 at 10 16 08 AM|

nickttng (approved):
Given the scope of the PR description, this looks good to me

zachmargolis (approved):

aduth:
Build failure is from visual regression, partially expected from the addition of the new "Button Groups" documentation preview page.

Looks like there's also some strange text encoding in one of the versions' content for accordions, hoping it's just a flakey issue.

image

PR #367, opened 2023-08-03
Upgrade to Ruby 3.1

aduth:
**Why?** Ruby 2.7 is EOL and unsupported by Cloud.gov Pages


jest-puppeteer.config.js
```diff
@@ -2,7 +2,7 @@ process.env.JEST_PORT = process.env.JEST_PORT || '40410';
 
 module.exports = {
   server: {
-    command: `ruby -run -ehttpd ./dist -p${process.env.JEST_PORT}`,
+    command: `bundle exec ruby -run -ehttpd ./dist -p${process.env.JEST_PORT}`,
```
aduth, line 5: Context on this change and the Webrick dependency:

Prior to Ruby 3, webrick was included by default. [This is no longer the case with Ruby 3.](https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/#:~:text=The%20following%20libraries%20are%20no%20longer%20bundled%20gems%20or%20standard%20libraries.%20Install%20the%20corresponding%20gems%20to%20use%20these%20features.) We use this script to create a simple file server, and [it depends on Webrick](https://github.com/ruby/un/blob/72d7494af7554ce2596e72b801e4a6aa203d3173/lib/un.rb#L333). So we now need to install Webrick as a dependency. Since we're doing that through Bundler, the command here changes to run Ruby through `bundle exec`, so that the Bundler-installed dependency is picked up.

.circleci/config.yml
```diff
@@ -25,7 +25,7 @@ commands:
     steps:
       - node-install
       - restore_cache:
-          key: dependency-cache-ruby-2-7-{{ checksum "package-lock.json" }}-{{ checksum "Gemfile.lock" }}
+          key: dependency-cache-ruby-3-1-{{ checksum "package-lock.json" }}-{{ checksum "Gemfile.lock" }}
```
aduth, line 28: Maybe this should checksum `.ruby-version` so we don't have to maintain it manually.

```suggestion
          key: dependency-cache-ruby-{{ checksum ".ruby-version" }}-{{ checksum "package-lock.json" }}-{{ checksum "Gemfile.lock" }}
```

.circleci/config.yml
```diff
@@ -36,7 +36,7 @@ commands:
           name: Install dependencies
           command: npm ci
       - save_cache:
-          key: dependency-cache-ruby-2-7-{{ checksum "package-lock.json" }}-{{ checksum "Gemfile.lock" }}
+          key: dependency-cache-ruby-3-1-{{ checksum "package-lock.json" }}-{{ checksum "Gemfile.lock" }}
```
aduth, line 39:
```suggestion
          key: dependency-cache-ruby-{{ checksum ".ruby-version" }}-{{ checksum "package-lock.json" }}-{{ checksum "Gemfile.lock" }}
```

charleyf (approved):
I got this running with ruby 3.1.4 and checked:
- It runs locally for me.
- The various `npm run ...` commands seem to work.

Thanks for doing this!

PR #369, opened 2023-08-14
LG-10523: Add Image Example Component

charleyf:
## 🎫 Ticket
https://cm-jira.usa.gov/browse/LG-10523

## 🛠 Summary of changes
This PR adds a new `image-example` component to the documentation and styles of the `identity-design-system`.

[Design reference link ](https://www.figma.com/file/amLP3Kk1PFRnm6Bk82PbxV/LG-10238%3A-Visuals-of-do's-and-don'ts-for-image-capture?type=design&node-id=182-1503&mode=design&t=9KtGUA3p3GEVmqvR-0)(from ticket).

## 🙋🏼‍♂️ Open Questions
1. ~~As far as I can tell, this would be the first component in the `identity-design-system` that doesn't directly inherit from a component in USWDS. I feel like that indicates a problem, but I'm not sure how to resolve. I think the change I had to make to `src/scss/styles.scss` is another indicator that something is wrong with this approach?~~
    - This is fine, per Andrew's answer below.
1. ~~I feel like there might be a better way to implement the simple classes I'm introducing like `text-green`?~~
    - There is, see Andrew's very thorough answer below about semantic styles.


src/scss/packages/image-example/src/_styles.scss
```diff
@@ -0,0 +1,26 @@
...
+}
+
+.image-example figure {
+  display: table;
+}
+
+.image-example figcaption {
+  display: table-caption;
+  caption-side: bottom;
+}
```
charleyf, line 26: This is the cleanest way I found to make the text wrap under the image correctly. To see what I mean, try wrapping this whole component in divs of various widths:
```html
<div style="width:200px" >
  <div class="image-example">
    ...
  </div>
</div>
```
And
```html
<div style="width:1200px" >
  <div class="image-example">
    ...
  </div>
</div>
```

Without these rules, the caption will overflow the image's width at the larger sizes.

aduth:
> 1. I feel like that indicates a problem, but I'm not sure how to resolve. I think the change I had to make to `src/scss/styles.scss` is another indicator that something is wrong with this approach?

I think we should aim to have LGDS look and feel like a drop-in replacement for USWDS, as far as being able to support the same package import semantics, including [direct-source references to `/src/styles`](https://designsystem.digital.gov/documentation/migration/#using-package-source).

In other words, I think how you've implemented it as far as folder structure is exactly how it should be, since `/src/styles` is our first-party styles.

src/scss/packages/image-example/src/_styles.scss
```diff
@@ -0,0 +1,26 @@
+@use 'uswds-core' as *;
+
+.image-example .bottom-border-green {
```
aduth, line 3: I think we should name classes using semantics, not colors. Similar to how alert banner varies on [states like success & error](https://designsystem.digital.gov/design-tokens/color/state-tokens/), not green and red.

I'd expect something like `.image-example--correct` and `.image-example--incorrect` as modifier classes, rather than green and red. I'd also expect the modifier should take place at the top-level element and the child elements would have standard class names, where we can customize appearance of those child elements using CSS nesting.

Example:

```html
<div class="image-example image-example--correct">
  <!-- ... -->
  <img src="image-example__image">
</div>
```

```scss
.image-example__image {
  // ...default image styles
}

.image-example--correct .image-example__image {
  border-bottom: units(1) solid color('success');
}

// Or:
// .image-example__image {
//   // ...default image styles
//
//   .image-example--correct & {
//     border-bottom: units(1) solid color('success');
//   }
// }
```

Reference:

- https://en.bem.info/methodology/naming-convention/#two-dashes-style
- https://en.bem.info/methodology/key-concepts/

charleyf: This was really helpful, thanks for writing all this up. I rewrote ([commit](https://github.com/18F/identity-design-system/pull/369/commits/f56f6fcc1250f6f11ce26eb5c1796ba2a58aa80c)) the styling to:
- Be semantic
- Be applied at the top level.
- Follow the `block--elem` naming convention linked.
- Use the `units(...)` and `color(...)` utilities.

I extracted the shared styles into the `@mixin image-example($color)` mixin as well. I'm not sure about this. It neatly encapsulates that the only difference between `correct` and `incorrect` is one color. On the other hand, it depends more heavily on sass functionality and might be harder to read.

docs/_components/image-example.md
```diff
@@ -0,0 +1,57 @@
...
+lead: An image example displays an image with a do/don't header to help users upload  images correctly.
+---
+
+This is not a component in the USWDS design system.
+
+## Image Example Do
+
+
+{% capture example %}
+<div class="image-example image-example--correct">
```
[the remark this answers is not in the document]
charleyf: Keeping it consistent makes sense to me, especially since that's what we did for the other custom components. I prepended `usa-` to all the style and file names ([commit](https://github.com/18F/identity-design-system/pull/369/commits/51100fca9ed723957325dcc1e7bb3947b93818ea)).

src/scss/styles.scss
```diff
@@ -1,2 +1,3 @@
 @use 'uswds-core';
 @forward 'uswds';
+@use 'image_example';
```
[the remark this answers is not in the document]
charleyf: > This would ensure it's included in the precompiled CSS version of the package, but it would not be included the [Sass entrypoint](https://github.com/18F/identity-design-system#styles), which should be expected to include all components of the design system.

After some research, I'm still not sure I'm fully understanding the difference here, but I've made that change ([commit](https://github.com/18F/identity-design-system/pull/369/commits/2c424274b745bf2c6280f488bce5b0c8fb06d139)).

I also prepended `usa-` ([commit](https://github.com/18F/identity-design-system/pull/369/commits/51100fca9ed723957325dcc1e7bb3947b93818ea)).

src/scss/packages/_uswds.scss
```diff
@@ -4,3 +4,4 @@
 // Custom Login.gov Design System components
 @forward 'usa-code/src/styles';
 @forward 'usa-verification-badge/src/styles';
+@forward 'usa-image-example';
```
aduth, line 7: Nit: Even just for consistency, I'd reference `/src/styles` directly like the others.

```suggestion
@forward 'usa-image-example/src/styles';
```

The difference being that the main package entrypoint would potentially have dependencies (it doesn't in our case), and loading those dependencies would be redundant from the main `uswds` package, since they're already expected to be included. The redundant loading can have a performance impact (see also https://github.com/uswds/uswds/pull/5241#issuecomment-1540686386).

aduth (approved):
Couple nitpicky comments, but it looks great to me! 👍

src/scss/styles.scss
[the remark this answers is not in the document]
charleyf: Thanks for spelling that out, this makes a lot more sense to me now.

src/scss/packages/_uswds.scss
charleyf: Got it. Making this change is basically telling sass to switch from:
- `@forward 'usa-image-example';` - Include the component styles and all dependencies.
- `@forward 'usa-image-example/src/styles';` - Include the component styles, but not dependencies.

aduth: Yep, exactly. If we needed dependencies, we could add them to `src/scss/packages/_usa-image-example.scss`†, and because this file `_uswds.scss` includes _everything_, we should be able to confidently assume we'd not need those dependencies to be reloaded, and therefore we can improve performance by referencing the styles directly.

----

† This is what dependencies look like in a USWDS component, for example:

https://github.com/uswds/uswds/blob/develop/packages/usa-accordion/_index.scss

docs/_components/image-example.md
```diff
@@ -0,0 +1,57 @@
...
+    <figcaption>
+      <ul class="usa-icon-list usa-icon-list--size-md">
+        <li class="usa-icon-list__item">
+          <div class="usa-icon-list__icon">
+            <svg class="usa-icon" aria-hidden="true" role="img">
+              <use xlink:href="{{ site.baseurl }}/assets/img/sprite.svg#cancel"></use>
+            </svg>
+          </div>
+          <div class="usa-icon-list__content">
+            <span class="image-example__do-dont">DON'T:</span> Use a white background, but make the text longer for testing the way the word wrap behaves.
```
aduth, line 49: More nit: Can we pick a consistent guidance for the space being within or after the closing `</span>` ? We've got one of each approach in these examples.

charleyf: Done! ([commit](https://github.com/18F/identity-design-system/pull/369/commits/852ed0b9ec308b4640e6048374fdad6a5fa4bc5e))

Side note: I'm planning on talking to design (Kelli) again tomorrow afternoon, then merging this if it meets her approval.

src/scss/packages/usa-image-example/src/_styles.scss
```diff
@@ -0,0 +1,43 @@
...
+  .img--bordered {
+    border-top: units(1px) solid color('base-lighter');
+    border-right: units(1px) solid color('base-lighter');
+    border-left: units(1px) solid color('base-lighter');
+  }
+}
+
+@mixin usa-image-example($color) {
+  img {
+    box-shadow: 0 units(1) 0 0 color($color);
```
charleyf, line 24: This was `border-bottom` but the combination of a thick `border-bottom` and a thin border on all the other sides made the `border-bottom` have tapered sides. This changes all the edges to be 90degrees.

src/scss/packages/usa-image-example/src/_styles.scss
charleyf, line 1: Oddly, there were no styles needed for the unmodified`.usa-image-example` class so I removed it.

src/scss/packages/usa-image-example/src/_styles.scss
```diff
@@ -0,0 +1,41 @@
...
+  display: table;
+}
+
+.usa-image-example__figcaption {
+  margin-top: units(1.5);
+  display: table-caption;
+  caption-side: bottom;
+}
+
+.usa-image-example__img--bordered {
```
aduth, line 29: As a BEM modifier, the `--bordered` suffix here implies to me that this is a variant on some base styles that don't include the bordering. But it's my understanding that we always want to have a border, or at least by default. So I'd suggest removing the modifier.

I also have a slight preference for no abbreviations, but I can also see a case being made for it matching the tag name.

```suggestion
.usa-image-example__image {
```

src/scss/packages/usa-image-example/src/_styles.scss
```diff
@@ -0,0 +1,41 @@
+@use 'uswds-core' as *;
+
+@mixin usa-image-example($color) {
+  img {
```
aduth, line 4: To the previous comment, should this be included in the `.usa-image-example__image` styles?

src/scss/packages/usa-image-example/src/_styles.scss
charleyf, line 29: When an image has a lighter background we want to lightly border the top, left, and right sides. We don't want this border when the image has a darker background.

For naming this, I tried to follow: https://en.bem.info/methodology/naming-convention/
Specifically, I looked at this example: `menu__item_visible` which in "two dashes style" ([link](https://en.bem.info/methodology/naming-convention/#two-dashes-style)) would end up being: `menu__item--visible` 

So I think I ended up with:
- block-name: `usa-image-example`
- elem-name: `image` (currently `img` but fixing now)
- mod-name: `bordered`

charleyf: Thanks this matches up well with what I'm thinking/intending. I just posted a review here, but:
- `--bordered` should imply that there's a variant. This is a variant (see comment below).
    - Since it's confusing maybe `--fully-bordered` ? Or some other name?
- I just renamed `img` to `image` that was a holdover from the tag name. Agreed it's better as `image`.

src/scss/packages/usa-image-example/src/_styles.scss
charleyf: Ah, missed those. I'll switch all of the mixins to use the correct classes.

charleyf: Done ([commit](https://github.com/18F/identity-design-system/pull/369/commits/db1511fe354aae3f53c25f15ef855cf2ebf52d56)).

src/scss/packages/usa-image-example/src/_styles.scss
charleyf: without--bordered
with--bordered

charleyf: Screenshots in this [comment](https://github.com/18F/identity-design-system/pull/369#discussion_r1303095251) (on this PR).

charleyf: The reason not to give every image a full border is it causes an odd little ledge with dark background images. Also, it's not necessary and might cause spacing issues down the line.
dark-background-with--bordered

aduth: Oh! I didn't realize we want to have the option to exclude the border. That makes total sense to have the modifier then. Especially now after you added the base class styles in db1511f.

aduth (approved):
>Thank you for the reviews @aduth! I learned a lot and will do my best to carry that knowledge forward to Timnit and my/our next PR in this repo.

Great work on this!

I also think this is a great candidate for the sorts of use-case-driven new components that are good to propose back to the main USWDS project for consideration. Especially since they already have [similar sort of "Do" / "Don't" content in their own documentation](https://designsystem.digital.gov/patterns/select-a-language/two-languages/#guidance).

PR #374, opened 2023-09-29
Disable Serif fonts

aduth:
**Why?** We don't currently use Serif fonts. Previously, we overrode the Serif font face configuration to set Public Sans. Instead, this proposes to _disable_ the serif font, and reconfigure the heading to use the Sans "role". This should be more semantically accurate, and optimizes the output to prevent [serif-related font utilities](https://designsystem.digital.gov/utilities/font-size-and-family/) from being output.

There should be no visual impact expected from these changes.

Related: https://github.com/18F/identity-idp/pull/9290

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Breaking Changes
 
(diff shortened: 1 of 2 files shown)


nickttng (approved):
LGTM!

PR #377, opened 2023-10-02
Mark font role variables as default

aduth:
Small revision to #374 prior to releasing 8.0.0 version. I noticed in testing the release an error in `identity-idp` when trying to override this value. As you can see in all surrounding variables, we usually set `!default` flag to allow downstream projects to reconfigure any value from the design system, so that's introduced here for consistency to the variables configured in #374.


zachmargolis (approved):

aduth:
>we usually set !default flag to allow downstream projects to reconfigure any value from the design system

As a follow-up, it'd be interesting to see if there might be a way to lint for this so that we're consistent.

PR #386, opened 2023-12-15
Replace Jest with Node built-in test runner

aduth:
## 🛠 Summary of changes

Replaces [Jest](https://jestjs.io/) as the build task runner with equivalent tests using [Node's new built-in test runner](https://nodejs.org/api/test.html).

**Why?**

- Simplifies, since we don't need all of what Jest includes
- Removes indirection through various Puppeteer abstractions
- Should improve performance of the test suite
   - Also includes changes to remove dependency on a fake DOM in `components` test
- Major reduction in dependency tree depth (see lines removed)
  - Helps maintainability by eliminating a common source of security advisories
    - This pull request mitigates from "2 moderate severity vulnerabilities" to "found 0 vulnerabilities"
- Node's built-in test runner looks and feels a lot more like [Mocha](https://mochajs.org/), the test library we use in `identity-idp`

## 📜 Testing Plan

Run `make test`


jest.config.js
```diff
@@ -1,20 +0,0 @@
...
-  notify: true,
-  preset: 'jest-puppeteer',
-  transform: {
-    '\\.js$': 'babel-jest',
-  },
-  testMatch: [
-    'ONLY_VISUAL_REGRESSION_TEST' in process.env
-      ? '**/test/screenshot.test.js'
-      : '**/test/**/*.test.js',
-  ],
```
aduth, line 15: Need to find a way to maintain this behavior (we split Visual Regression to a separate job in CircleCI).

test/accessibility.test.mjs
aduth, line 1: Unfortunately Git/GitHub didn't preserve these as a file revision. These specs are largely the same with a few differences around setup:

- Replace Jest's `test` blocks with equivalent `describe` and `it`
- Specs which rely on Puppeteer now set up the server in `before` / `after` blocks
- Reference updated root URL for static file server, since we don't have `JEST_PORT` environment variable
   - Instead, use the address of the local server variable
- Convert CommonJS to ESM
   - Including equivalents for features not supported in ESM, e.g. `__dirname`

aduth:
Some one-to-one comparisons on test timing (only single test run):

- "Run visual regression test"
  - [Before](https://app.circleci.com/pipelines/github/18F/identity-design-system/1308/workflows/28e93e39-b6aa-4c1d-9f7b-b6fc8cfa2858/jobs/3616): 1m16s
  - [After](https://app.circleci.com/pipelines/github/18F/identity-design-system/1312/workflows/aac294e2-a8b0-4128-a6e1-519f0510e21b/jobs/3630): 56s
- "Run integration tests"
  - [Before](https://app.circleci.com/pipelines/github/18F/identity-design-system/1308/workflows/28e93e39-b6aa-4c1d-9f7b-b6fc8cfa2858/jobs/3617): 24s
  - [After](https://app.circleci.com/pipelines/github/18F/identity-design-system/1312/workflows/aac294e2-a8b0-4128-a6e1-519f0510e21b/jobs/3628): 16s

I could probably optimize this even further, since it looks like each test is doing a duplicated rebuild of the documentation site assets within the test command:

- Documentation site assets may not be needed at all for the non-visual regression test
- Site assets are needed for the visual regression test, but should only be run once

aduth:
becafc3 removes the separate "Build assets and site" step which should effectively eliminate around 7s from the overall build time for each.

jest.config.js
aduth: Restored in 7c3b564.

zachmargolis (approved):

PR #387, opened 2023-12-18
Replace Babel with ESBuild

aduth:
## 🛠 Summary of changes

Replaces/consolidates [Babel](https://babeljs.io/) as the JavaScript code transpiler with [esbuild](https://esbuild.github.io/).

**Why?**

- Consolidates and ensure consistency with existing ESBuild usage for documentation / test build
- It's faster, thereby shortening feedback loop in local development and CI build runtime
- Reduces dependency tree depth (see lines removed), which helps maintainability by eliminating a common source of security advisories
- ESBuild is well-aligned to the project's goals, targeting semi-modern browsers with bundling features, and support for potential future shift to TypeScript

## 📜 Testing Plan

1. `npm install`
2. `npm run build:pkg`
3. Observe local `build/esm` and `build/cjs` folders include transpiled JavaScript in respective ESM and CommonJS formats


.browserslistrc
```diff
@@ -1 +1 @@
-> 1% or IE 11
+> 0.5%
```
aduth, line 1: This change is necessary since ESBuild doesn't support all transpilations for IE11. 

Admittedly, I thought we'd long dropped support for IE11 in this project, since we already had done so in our other projects.

Technically this is a breaking change, so we'd probably want to note it in the CHANGELOG and bump the major version if we move forward with it.

aduth:
I think I'd like to do this, but I'll hold on it for now 'til we have other breaking changes to release.

It may end up having a nice downstream effect of optimizing bundle sizes, since the transpiled output currently may be inflating the size due to current support for IE11.

aduth:
Reviving this as there's plans for a new major version, which we can lump this breaking change into.

See https://github.com/18F/identity-design-system/pull/390#issuecomment-1883779930

scripts/esbuild-targets.mjs
```diff
@@ -0,0 +1,44 @@
...
+    // [["chrome", "119"], ["chrome", "118"], ...] ➡️ [["chrome", "118"] ...]
+    .filter(([name, version], _index, array) =>
+      array.every(([otherName, otherVersion]) => otherName !== name || version <= otherVersion),
+    )
+    // Convert back to ESBuild string target format
+    // [["chrome", "118"], ...] ➡️ ["chrome118" ...]
+    .map(([name, version]) => [name, version].join(''))
+    // Convert to comma separated list
+    // ["chrome118", "safari15.6" ...] ➡️ "chrome118,safari15.6"
+    .join(),
```
zachmargolis, line 43: the comments are great! if this was changed often we might want to run unit tests for it?

zachmargolis (approved):

scripts/esbuild-targets.mjs
aduth: Yeah, I suspect that'd be helpful, though there's not a great pattern for testing the logic of a file like this. Ideally we'd move the transformation code into its own package, and I do think component-specific packages (like USWDS's [packages](https://github.com/uswds/uswds/tree/develop/packages) pseudo-monorepo) makes sense for the design system.

Another option could be to use an existing package that does this already, like [`browserslist-to-esbuild`](https://github.com/marcofugaro/browserslist-to-esbuild) or [`esbuild-plugin-browserslist`](https://github.com/nihalgonsalves/esbuild-plugin-browserslist), though AFAICT neither of those can be used via the CLI.

PR #400, opened 2024-01-26
Use fully-qualified path for components

aduth:
## 🛠 Summary of changes

Fixes a regression in #395 where using ES Modules requires imports to reference the full path of any file.

See: https://nodejs.org/api/esm.html#mandatory-file-extensions

Error after upgrading to `8.1.0` in `identity-site`:

```
ERROR in ./node_modules/@18f/identity-design-system/build/esm/index.js 1:0-29
Module not found: Error: Can't resolve './components' in '/Code/identity-site/node_modules/@18f/identity-design-system/build/esm'
Did you mean 'index.js'?
BREAKING CHANGE: The request './components' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

This will be planned to be released as a 8.1.1 patch release, cherry-picked onto a branch from the v8.1.0 tag.

## 📜 Testing Plan

While I'm pretty sure I had tried to do this already, you could try updating `identity-site` `package.json` to reference its dependency as a file path, run `npm install`, and verify that `npm run build:js` completes successfully

```
"@18f/identity-design-system": "file:../identity-design-system",
```


zachmargolis (approved):

aduth:
Visual regression failure doesn't appear to be related.

PR #402, opened 2024-01-29
Use Incognito in visual regressions to avoid visited link styling

aduth:
## 🛠 Summary of changes

Attempts to fix some test flakiness in visual regression specs, where links may sometimes be "visited" and appear in a purple color depending on test parallelism, since pages are now run concurrently across multiple tabs as of #389.

The solution is to use a Chrome Incognito (private browsing) context, where history will not be remembered. In my brief testing of a live Incognito tab, links will never show with visited styling.

## 📜 Testing Plan

Build should pass.

Test locally:

```
ONLY_VISUAL_REGRESSION_TEST=1 make test
```

### DIFF
--- a/docs/_components/prose.md
+++ b/docs/_components/prose.md
@@ -45,9 +45,7 @@ This is an unordered list
 
 This is a [text link](?{{ site.time | date: "%s%N" }}) on a light background, using the LDGS primary-blue color
 
-This is a <a href="{{ site.baseurl }}" class="usa-link usa-link--visited">visited text link</a>, using the USWDS setting (#54278F)
-
-This is a link that goes to an <a href="{{ site.baseurl }}" class="usa-link usa-link--external">external website</a>.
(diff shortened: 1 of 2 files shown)


aduth:
Reverting to draft since this did not resolve the test failures.

aduth:
I think the initial changes make sense, but the bigger issue is that `usa-link--visited` doesn't exist, and there's no real way to "force" a link to be styled as visited. The way this is done in [USWDS documentation](https://designsystem.digital.gov/components/link/) is with a custom CSS class `.usa-color-text-visited` which is specific to their documentation site ([source](https://github.com/uswds/uswds-site/blob/8358991588381210f51afedfef8a527b46108a4f/css/_uswds-theme-custom-styles.scss#L843-L845)). We don't really have a good way to replicate that, so I opted to just remove it for now, and add a cache busting query parameter on other links which may be considered "visited".

The build will fail since the content of the page is being changed from the live site, but it should pass moving forward I hope.

zachmargolis (approved):

PR #403, opened 2024-02-02
Run visual regression testing as self-contained without external hosting

aduth:
## 🛠 Summary of changes

Updates visual regression testing to avoid reliance on a hosted copy of the "main" branch, in anticipation that we may need to decommission current hosted references.

## 📜 Testing Plan

Build should pass.

Test locally by generating "snapshot" on each of `main` and this branch (`aduth-internal-visual-regression`) before running visual regression test:

1. `git checkout main`
2. `git checkout aduth-internal-visual-regression -- scripts/snapshot.js`
3. `scripts/snapshot.js`
4. `rm scripts/snapshot.js`
5. `git checkout aduth-internal-visual-regression`
6. `scripts/snapshot.js`
7. `ONLY_VISUAL_REGRESSION_TEST=true make test`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,6 +101,8 @@ jobs:
     steps:
       - checkout
       - bundle-npm-install
+      - attach_workspace:
+          at: tmp/screenshot/branches
       - run:
           name: Run visual regression test
           command: npm test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,10 +102,10 @@ jobs:
           name: Run visual regression test
           command: npm test
       - store_artifacts:
-          path: tmp/results
+          path: tmp/screenshot/diff
           destination: results
       - store_test_results:
-          path: tmp/results
+          path: tmp/screenshot/diff
 
 ignore_main: &ignore_main
   filters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ commands:
       - persist_to_workspace:
           root: tmp/screenshot/branches
           paths:
-            - *
+            - '*'
 
 jobs:
   lints:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,9 +77,10 @@ jobs:
   snapshot-main:
     executor: ruby_browsers
     steps:
+      - checkout
       - run:
           name: Checkout main
-          command: git clone -b main << pipeline.trigger_parameters.github_app.repo_url >>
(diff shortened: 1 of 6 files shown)


test/screenshot.test.js
```diff
@@ -80,41 +45,14 @@ function fillImageToSize(image, width, height) {
...
   paths.forEach((path) => {
     test(path, async () => {
-      const localURL = `http://localhost:${port}`;
-      const page = await browserContext.newPage();
-      const local = await getScreenshot(page, localURL + path);
-      const remote = await getScreenshot(page, REMOTE_HOST + path);
-      const localPNG = PNG.sync.read(local);
-      const remotePNG = PNG.sync.read(remote);
+      const localPNG = PNG.sync.read(await readFile(join(BRANCH_SNAPSHOTS_DIRECTORY, path)));
+      const remotePNG = PNG.sync.read(await readFile(join(MAIN_SNAPSHOTS_DIRECTORY, path)));
```
zachmargolis, line 55: now that now that these are both "local" should we call it "branch" and "main"?
```suggestion
      const branchPNG = PNG.sync.read(await readFile(join(BRANCH_SNAPSHOTS_DIRECTORY, path)));
      const mainPNG = PNG.sync.read(await readFile(join(MAIN_SNAPSHOTS_DIRECTORY, path)));
```

zachmargolis (approved):

.circleci/config.yml
```diff
@@ -81,10 +105,26 @@ jobs:
       - store_test_results:
           path: tmp/results
 
+ignore_main:
```
zachmargolis, line 108: I think you need to name this reference in order to splat it below?
```suggestion
ignore_main: &ignore_main
```

aduth: Thanks, was just looking at that and was rather confused 😅

test/screenshot.test.js
aduth: Yep, good call, I think I'd meant to circle back to that but forgot.

aduth: Updated in e7b889d.

aduth:
You can see this is working by looking at "Visual Regression" result for the commit ef27789e9c46fd5392b28dd39d9a6c91af5156d2.

The test fails and the diff is uploaded under the "Artifacts" tab:

```
✖ accordions.png (3789.801049ms)
  AssertionError [ERR_ASSERTION]: Expected "accordions.png" to visually match the live site.
```

image

PR #404, opened 2024-02-02
Remove references to Cloud.gov Pages branch previews

aduth:
## 🛠 Summary of changes

Removes all references to Cloud.gov Pages, including documented workflow for using branch previews.

This is being done in preparation to delete the Cloud.gov Pages site altogether, due to upcoming ATO requirements.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,9 +81,6 @@ jobs:
       - run:
           name: Checkout main
           command: git checkout main
-      - run:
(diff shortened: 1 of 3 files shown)


aduth:
Unclear the issue with the visual regression test. It looks like one of the screenshots got partly clipped. Especially curious timing with the merge of #403, though I can't imagine what would have changed to produce that effect. Will keep an eye on it for possible improvements.

PR #405, opened 2024-02-05
Remove raw color utilities

aduth:
## 🛠 Summary of changes

Updates theme configuration to remove color palettes for anything other than "required", "theme", or "state", effectively removing ad hoc color usage (e.g. green) in favor of semantic color tokens (e.g. success).

**Why?**

- In almost all cases, the state colors are more semantic in representing the intent of the color
- The colors were wrong, since we weren't customizing the underlying [system colors](https://designsystem.digital.gov/design-tokens/color/system-tokens/), e.g. `text-green` was `#538200` vs. `text-success` as `#18852e`
   - This was noticed by @Sgtpluck in https://github.com/18F/identity-dev-docs/pull/449
- Slight reduction in stylesheet size

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,35 @@
 
 ### Breaking Changes
 
-- [Prose](https://designsystem.digital.gov/components/prose/) styling overrides have been removed. Text formatting should now adhere to the defaults from the U.S. Web Design System.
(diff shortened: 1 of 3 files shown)


Sgtpluck (approved):

zachmargolis (approved):

aduth:
The visual regression tests caught a regression, which was fixed in 7461411. I suspect the tests will still fail due to slight color differences between `text-green`/`text-red` and `text-success`/`text-error`.

PR #411, opened 2024-02-28
Apply SVG optimization recursively to image subfolders

aduth:
## 🛠 Summary of changes

Fixes an issue where the SVG optimization was only applying to images within the top-level `src/img` directory, and not nested subdirectories (e.g. `src/img/alerts`, `src/img/favicons`, `src/img/illustrations`).

This fixes that issue by specifying the recursive option `-r` to the command. It also updates the tool and re-optimizes images, in order to avoid regressions where earlier versions of the tool would actually inflate existing sizes of some of these images.

(Temporarily merges to #410, where this issue was identified, and where a number of images are being removed)

## 📜 Testing Plan

Verify that running the optimization lint task passes successfully.

```
npm install
make lint-optimized-assets
```


Makefile
```diff
@@ -29,7 +29,7 @@ typecheck:
 	npm run typecheck
 
 optimize-svg:
-	$(NODE_BIN)/svgo --config svgo.config.js -f src/img
+	$(NODE_BIN)/svgo --config svgo.config.js -r -f src/img
```
zachmargolis, line 32: one tip I've seen is to always use long flag names when things are checked in to source so they document themselves better
```suggestion
	$(NODE_BIN)/svgo --config svgo.config.js --recursive --folder src/img
```

zachmargolis (approved):
LGTM

Makefile
aduth: Yeah, that's a good point 👍 Will update and keep that in mind for the future.

aduth:
Oops, I'd planned to merge #412 to `main` independently from this one, but forgot I'd chained them together, so it got merged here. No big deal, just a little extra noise on the changes.

PR #413, opened 2024-03-01
Add TypeScript types for JavaScript components

aduth:
## 🛠 Summary of changes

Includes TypeScript types in the published NPM package for JavaScript components.

This basically republishes `@uswds/uswds` DefinitelyTyped types, with corrected LGDS aliases and missing type definitions for newer components. There were a handful of different errors with the DefinitelyTyped types, and I may work to propose an upstream patch to resolve these, which would hopefully allow the typings of this package to be simplified.

This doesn't include enforcement that the typings exactly reflect the exported components, which I'd like to add in future iterations to provide similar guarantees to what's asserted in [`test/components/index.test.js`](https://github.com/18F/identity-design-system/blob/main/test/components/index.test.js).

## 📜 Testing Plan

I tested this in `identity-idp` by changing its `@18f/identity-design-system` dependency to point to the local filesystem copy and verifying that component references include autocompletion and type validation.

```diff
diff --git a/package.json b/package.json
index b6af77174..be85052f8 100644
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "build:css": "build-sass app/assets/stylesheets/*.css.scss app/components/*.scss --load-path=app/assets/stylesheets --out-dir=app/assets/builds"
   },
   "dependencies": {
-    "@18f/identity-design-system": "^8.1.2",
+    "@18f/identity-design-system": "file:../identity-design-system",
     "@babel/core": "^7.20.7",
     "@babel/preset-env": "^7.15.6",
     "@babel/preset-react": "^7.14.5",
```

In `identity-idp`:

1. Apply patch above
2. Clear and reinstall dependencies: `rm -rf node_modules && yarn`
3. Run TypeCheck: `yarn typecheck`
4. (Optional) Check auto-completion support in supported IDE such as VS Code

Example (`app/javascript/packages/clipboard-button/clipboard-button-element.ts`):

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,10 @@
   - `close.svg` (use `close` icon instead)
   - `hero.png`
 
+### New Features
+
+- Add TypeScript types for JavaScript components.
+
 ### Bug Fixes
 
 - Add missing JavaScript exports for `button`, `inPageNavigation`, `inputMask`, `languageSelector`, and `range`. ([#407](https://github.com/18F/identity-design-system/pull/407))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@
 
 ### New Features
 
-- Add TypeScript types for JavaScript components.
+- Add TypeScript types for JavaScript components. ([#413](https://github.com/18F/identity-design-system/pull/413))
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


zachmargolis (approved):

aduth:
> There were a handful of different errors with the DefinitelyTyped types, and I may work to propose an upstream patch to resolve these, which would hopefully allow the typings of this package to be simplified.

Upstream: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68863

PR #414, opened 2024-03-01
Align JavaScript component names with USWDS

aduth:
## 🛠 Summary of changes

Updates JavaScript components to ensure that their names match those exported by USWDS. This is an inconsistency within USWDS itself between its package directory names and the corresponding exported JavaScript component name.

**Why?**

- This package is meant to operate as a drop-in replacement for USWDS, so its exported members should match 1-to-1

## 📜 Testing Plan

Verify the modified test passes:

```
npm run build
node --test test/components/index.test.js
```

### DIFF
--- a/test/components/index.test.js
+++ b/test/components/index.test.js
@@ -16,7 +16,7 @@ describe('components', () => {
       uswdsComponentsIndex.match(/module\.exports = \{([\s\S]+)\};/)
     );
     const uswdsComponentKeys = Array.from(exports.matchAll(/(?:\s+([a-zA-Z]+)),/g))
-      .map((match) => match[1])
+      .map(([, capturedKey]) => capturedKey)
       .sort();
 
     assert(uswdsComponentKeys.length);
--- a/test/components/index.test.js
+++ b/test/components/index.test.js
@@ -1,17 +1,6 @@
 import { describe, it } from 'node:test';
 import assert from 'node:assert';
 import { readFile } from 'node:fs/promises';
-import { basename, dirname } from 'node:path';
-import glob from 'fast-glob';
-
-/**
- * Converts a string to camel case.
- *
- * @param {string} str
- *
- * @return {string}
- */
-const camelCase = (str) => str.replace(/-([a-z])/g, (_match, character) => character.toUpperCase());
 
 describe('components', () => {
   it('re-exports all uswds components', async () => {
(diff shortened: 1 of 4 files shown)


test/components/index.test.js
```diff
@@ -1,27 +1,22 @@
...
-      .sync('./node_modules/@uswds/uswds/packages/*/src/index.js')
-      .map((path) => camelCase(basename(dirname(dirname(path))).replace(/^_?usa-/, '')))
+    const uswdsComponentsIndex = await readFile(
+      'node_modules/@uswds/uswds/packages/uswds-core/src/js/index.js',
+      'utf-8',
+    );
+    const [, exports] = /** @type {RegExpMatchArray} */ (
+      uswdsComponentsIndex.match(/module\.exports = \{([\s\S]+)\};/)
+    );
+    const uswdsComponentKeys = Array.from(exports.matchAll(/(?:\s+([a-zA-Z]+)),/g))
```
aduth, line 18: I considered pulling in a project like [`cjs-module-lexer`](https://www.npmjs.com/package/cjs-module-lexer) for this, though we already have some regular expressions above for parsing the ESM from this project.

test/components/index.test.js
```diff
@@ -1,27 +1,22 @@
...
+      .map((match) => match[1])
```
zachmargolis, line 19: can we destructure the regex match?
```suggestion
      .map(([, capturedKey]) => capturedKey)
```

zachmargolis (approved):

test/components/index.test.js
aduth: We can!

aduth:
Sad thing is I discovered this is the inverse of a breaking change we previously implemented in 7.0.0 🙈 

https://github.com/18F/identity-design-system/pull/321/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

Hopefully this'll stick. I don't love that the names are inconsistent with package directory in USWDS (which is where the names [`header`](https://github.com/uswds/uswds/tree/develop/packages/usa-header) and [`validation`](https://github.com/uswds/uswds/tree/develop/packages/uswds-validation) come from), but I think I'd rather be consistent with how USWDS exposes these in their [main index](https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/js/index.js).

PR #390, opened 2023-12-19
Reset and document Prose component styles

aduth:
## 🛠 Summary of changes

Removes override styles for the [Prose component](https://designsystem.digital.gov/components/prose/) and adds a preview page to the internal documentation.

Related Slack discussion: https://gsa-tts.slack.com/archives/C01JVKYE4KD/p1702914619083849

Related LGDS Office Hours discussion notes: https://docs.google.com/document/d/10b7Gis_VxYT3roR01MX4hwPa6-Tq29YBaLPIB3vWnUw/edit#bookmark=id.w558e3npl6ym

Live preview URL: https://federalist-340d8801-aa16-4df5-ad22-b1e3e731098e.sites.pages.cloud.gov/preview/18f/identity-design-system/aduth-typography-prose/prose/

(Draft pending audit of effect of removing overrides)

## 👀 Screenshots

federalist-340d8801-aa16-4df5-ad22-b1e3e731098e sites pages cloud gov_preview_18f_identity-design-system_aduth-typography-prose_prose_(Computer Standard)

PR #415, opened 2024-03-01
Fix direct imports to prose styles

aduth:
## 🛠 Summary of changes

Regression of #390

Fixes an issue where code directly importing `usa-prose` would fail due to missing `_overrides.scss` file removed in #390.

```
Exception [Error]: Error: Can't find stylesheet to import.
  ╷
2 │ @forward './overrides';
  │ ^^^^^^^^^^^^^^^^^^^^^^
```

The project's current build pipeline would only catch errors occurring in a full build. TBD I need to more fully understand how this doesn't cause a failure within this project, but the short of it seems to be that [this file](https://github.com/18F/identity-design-system/blob/main/src/scss/packages/usa-prose/src/_styles.scss) isn't being loaded.

## 📜 Testing Plan

I reproduced the issue on `main` by testing in `identity-idp`.

In `identity-idp`:

1. Change `@18f/identity-design-system` dependency in `package.json` to point to `file:../identity-design-system`
2. `rm -rf node_modules`
3. `yarn`
4. `yarn buld:css`

Before: Fails with error mentioned above
After: Succeeds

### DIFF
--- a/src/scss/packages/_usa-prose.scss
+++ b/src/scss/packages/_usa-prose.scss
@@ -1,1 +1,0 @@
-@forward 'usa-prose/index';
--- a/src/scss/packages/usa-prose/src/_styles.scss
+++ b/src/scss/packages/usa-prose/src/_styles.scss
@@ -1,2 +1,0 @@
-@forward 'usa-prose/src/styles/index';
-@forward './overrides';


zachmargolis (approved):

aduth:
Visual regression is expected since removal of overrides file causes "View Login.gov SCSS on GitHub" link to be removed from internal documentation.

image

PR #416, opened 2024-03-02
Update USWDS typings to incorporate corrections

aduth:
## 🛠 Summary of changes

Updates `@types/uswds__uswds` to the latest version to avoid needing to patch corrections for incorrect types, which are now fixed upstream.

This is a follow-up to https://github.com/18F/identity-design-system/pull/413#issuecomment-1973757382

## 📜 Testing Plan

```
npm run typecheck
```


index.d.ts
```diff
@@ -1,67 +1 @@
...
-export const navigation: typeof baseNavigation & BaseComponent;
-export const password: typeof basePassword & BaseComponent;
-export const range: Range;
-export const search: typeof baseSearch & BaseComponent;
-export const skipnav: typeof baseSkipnav & BaseComponent;
-export const timePicker: typeof baseTimePicker & BaseComponent;
-export const table: typeof baseTable & BaseComponent;
-export const tooltip: Tooltip;
-export const validator: typeof baseValidator & BaseComponent;
+export * from '@uswds/uswds/js';
```
zachmargolis, line 1: much simpler, glad this worked!

zachmargolis (approved):

PR #418, opened 2024-03-05
Add header documentation with mobile preview

aduth:
## 🛠 Summary of changes

This adds a new documentation page preview for the Header component, in response to a bug which surfaced in https://github.com/18F/identity-dashboard/pull/748#issuecomment-1977518011 related to active hover states for dropdown navigation links.

This doesn't actually fix the issue addressed in the comment above, and instead is limited to just adding a page to the internal documentation.

Notably, this includes a live, simulated mobile preview ~using an `iframe` receiver communicated using `postMessage`. For the security-minded, this should probably raise some hairs, though (a) there is protection to ensure messages are only received from the same domain, (b) this site is not publicly hosted anywhere at the moment, and (c) the site does not store any valuable data which would be worth targeting.~ **Edit:** After ace14d0, this was updated to use a simple static page as the frame source.

## 📜 Testing Plan

1. Locally, run `npm start`
2. Visit http://localhost:4000/header/

## 👀 Screenshots

Example mobile preview:

image


docs/_includes/helpers/mobile-embed.html
```diff
@@ -0,0 +1,39 @@
+
+<div class="mobile-embed">
```
zachmargolis, line 2: remove this extra blank line?
```suggestion
<div class="mobile-embed">
```

docs/mobile-embed-frame.md
```diff
@@ -0,0 +1,38 @@
...
+    return;
+  }
+
+  const preview = document.getElementById('preview');
+  preview.innerHTML = markup;
+  preview.querySelectorAll('script').forEach((script) => {
+    const replacement = document.createElement('script');
+    replacement.innerHTML = script.innerHTML;
+    script.parentNode.replaceChild(replacement, script);
+  });
```
zachmargolis, line 27: I know your [original comment mentioned this PR should raise some security eyebrows](https://github.com/18F/identity-design-system/pull/418#:~:text=For%20the%20security%2Dminded%2C%20this%20should%20probably%20raise%20some%20hairs)... just wanted to confirm yes eyebrows raised

zachmargolis (approved):

docs/mobile-embed-frame.md
aduth: Yeah, on this code in particular, it's strange that [browsers don't evaluate dynamically inserted script tags](https://stackoverflow.com/questions/1197575/can-scripts-be-inserted-with-innerhtml), while seemingly having no issue with inserting something equally exploitable like ``.

An alternative approach could be to have just a standalone HTML page that already has the component markup in it, though that might end up with a lot more duplication, especially if we wanted to use it for more components.

I might time-box and see if I can put something together for that and what it ends up looking like. I tried in an earlier iteration and had some trouble, but I think I was fighting more with Jekyll than the approach itself.

docs/_includes/helpers/mobile-embed.html
aduth: Yep good catch 👍

docs/mobile-embed-frame.md
aduth: For what it's worth, [I've implemented something like this in the past](https://github.com/Automattic/wp-calypso/pull/4619) and had previously spent a lot of time verifying the concept as safe, even with untrusted content in a secure context.

Although now that I write this, I probably need to harden this a bit more for the implementation to be _truly safe_ (specifically [`iframe` `sandbox` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#sandbox)). I'll see where I end up with my timeboxed experimentation whether this is an approach we go with.

aduth: Updated in ace14d0 to avoid the `postMessage` communication and dynamic script insertion, and instead using a simple static page for the frame.

aduth:
Visual regression failure is expected, since we're adding a new link to the sidebar navigation.

PR #422, opened 2024-03-06
Fix error when importing from ESM auto entrypoint

aduth:
## 🛠 Summary of changes

Fixes an error which occurs when trying to load the `auto` entrypoint in a bundler which resolves ESM imports.

```
Module not found: Error: Can't resolve './components' in '/node_modules/@18f/identity-design-system/build/esm'
```

This is an identical issue as what was addressed in #400, but the earlier pull request only addressed the instance of the issue within `index.js`. This pull request applies the same fix to `auto.js`.

This issue was surfaced as part of ongoing work in https://github.com/18F/identity-dashboard/pull/748

## 📜 Testing Plan

Verify included test passes:

1. `make build`
2. `node --test test/imports.test.js`

Optionally, verify in an external project that you can reference `/auto`:

1. In `identity-design-system`, ensure everything is built: `make build`
2. In external project, modify `package.json` version for `@18f/identity-design-system` to point to `file:../identity-design-system` (assumes `identity-design-system` is checked out to the parent folder)
3. In external project's JavaScript entrypoint that references the `@18f/identity-design-system`, import as: `import '@18f/identity-design-system/auto';`
4. Verify you have no errors when building JavaScript

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@
 
 - Add missing JavaScript exports for `button`, `inPageNavigation`, `inputMask`, `languageSelector`, and `range`. ([#407](https://github.com/18F/identity-design-system/pull/407))
 - Fix `dist/assets/js/main.js` compilation to properly consider browser support. ([#421](https://github.com/18F/identity-design-system/pull/421))
+- Fix strict ES Module import errors due to lack of fully-qualified file path when importing `auto` entrypoint. ([#422](https://github.com/18F/identity-design-system/pull/422))
 
 ### Internal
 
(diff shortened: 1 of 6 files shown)


zachmargolis (approved):
LGTM, neat test!

aduth:
> LGTM, neat test!

Yeah unfortunately it seems like TypeScript isn't taking nicely to the types in the `happy-dom` package 🙈 So might have to find another way to stub the DOM.

aduth:
> > LGTM, neat test!
> 
> Yeah unfortunately it seems like TypeScript isn't taking nicely to the types in the `happy-dom` package 🙈 So might have to find another way to stub the DOM.

Found a nice solution to this in b13d216. More detail in the commit message.

Bonus is that it might make TypeScript check a bit faster! (..."at the expense of type-system accuracy" 🙉 )

PR #423, opened 2024-03-07
Remove hidden attributes from header component code

aduth:
## 🛠 Summary of changes

Updates header component example source code to exclude `hidden` attributes.

These examples are meant to serve as reference implementations and therefore should be accurate.

The `hidden` attribute should not be included so that the menu sections appears as expanded on devices where JavaScript is disabled. The design system JavaScript will dynamically add back the `hidden` attribute, and when used in combination with the `init.js` initializer script, no flashing content should occur.

## 📜 Testing Plan

Verify there is no regression in the appearance of section navigation menus in the Header component example.

1. `npm start`
2. Go to http://localhost:4000/header/

### DIFF
--- a/docs/_components/header.md
+++ b/docs/_components/header.md
@@ -55,7 +55,7 @@ lead: >
             >
               <span>Section</span>
             </button>
-            <ul id="nav-section" class="usa-nav__submenu" hidden>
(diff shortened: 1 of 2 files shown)


aduth:
The visual regression test is catching the change in content in the page:

image

PR #430, opened 2024-03-19
Remove nav container from header component example

aduth:
## 🛠 Summary of changes

Improves header component markup example to avoid `<div class="usa-nav-container">`. It also updates the logo link to always point to the root regardless of `{{ site.baseurl }}` value (previously it would often link to the same page).

**Why?**

- This markup was originally sourced from `identity-site`, but the [default USWDS component guidance](https://designsystem.digital.gov/components/header/) for extended header does not include it
- The `usa-nav-container` creates additional horizontal padding that causes it to be misaligned with page content that includes navigation

## 📜 Testing Plan

Easiest to review with whitespace changes hidden, since the changes amount to removing a single wrapper element `<div class="usa-nav-container">`.

1. `npm start`
2. Visit http://localhost:4000/banner/
3. Observe reduced horizontal padding on header

## 👀 Screenshots

Before|After
---|---
image|image

### DIFF
--- a/docs/_components/header.md
+++ b/docs/_components/header.md
@@ -13,7 +13,7 @@ lead: >
 <div class="usa-header usa-header--extended">
   <div class="usa-navbar">
     <div class="usa-logo">
-      <a href="{{ site.baseurl }}">
+      <a href="{{ site.baseurl }}/">
         <img
           src="{{ site.baseurl }}/assets/img/login-gov-logo.svg"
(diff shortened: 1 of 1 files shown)


zachmargolis (approved):

nickttng (approved):

aduth:
Visual regression failure is catching the expected diff on the header appearance:

image

PR #431, opened 2024-03-19
Simplify and generalize navigation

aduth:
## 🛠 Summary of changes

Updates how side navigation is generated within the local documentation site, to avoid tying it specifically to components. 

This is part of a series of changes looking to introduce new sections for "Utilities" in addition to components, which will need to have its own side navigation.

## 📜 Testing Plan

The only meaningful change is that visiting the root URL will redirect automatically to the first component.

1. `npm start`
2. Visit http://localhost:4000/
3. Observe that you are redirected to http://localhost:4000/accordions/


docs/_plugins/resolve_permalink.rb
```diff
@@ -0,0 +1,29 @@
+module ResolvePermalink
+  # Resolve a permalink URI into a page or collection document:
+  #
+  # {{ '/' | resolve_permalink %} => home page
+  # {{ '/components/foo/' | resolve_permalink %} => from a collection
+  def resolve_permalink(href)
```
aduth, line 6: This implementation is not new, it already existed in the now-removed `uswds.rb`. This is the only utility from that module we still needed, and the naming "USWDS" was confusing, so I renamed the file and module to reflect the single remaining helper.

zachmargolis (approved):

aduth:
Visual regression tests caught a change that I wasn't intending, which is that previously the side navigation would "expand" and show sub-headings of a page, e.g. "Button sizes and states" on the Buttons page. This no longer happens. I think I'm fine with this side-effect and we could consider the [In-Page Navigation component](https://designsystem.digital.gov/components/in-page-navigation/) if we wanted something like this in the future, but good to highlight another instance of the visual regression tests doing their job.

Before|After
---|---
image|image

PR #434, opened 2024-03-19
Hide collapsed Accordion component content while JavaScript loads

aduth:
## 🛠 Summary of changes

This enhances accordion styles to appear hidden while JavaScript is loaded if expected to be collapsed once JavaScript finishes loading, avoiding the appearance of flashing content.

This backports changes added to the brochure site in https://github.com/GSA-TTS/identity-site/pull/1224

It also incorporates a number of improvements:

- Fixes an issue with https://github.com/GSA-TTS/identity-site/pull/1224 where it would not apply to sibling content of accordions where the button is wrapped inside a header container
- Inverts the selector to look for `:not([aria-expanded='true'])` instead of `[aria-expanded='false']` to support expected behavior where the `aria-expanded` attribute is not explicitly specified
- Updates component examples for Accordion:
   - Document differences between "Default" and "Bordered" variant options
   - Remove an unnecessary and useless `usa-accordion__container` container element which is not standard to USWDS and would make the JavaScript hiding behavior ineffective
   - Remove unused and unnecessary `data-test` attribute
   - Encourage use of explicit `aria-expanded="false"`
- Micro-optimizes `src/js/init.ts` to remove unnecessary curly braces on function body, syncing with upstream https://github.com/uswds/uswds/pull/5819

## 📜 Testing Plan

1. `npm start`
2. Visit http://localhost:4000/accordions/
3. Hard refresh
4. Observe no flashing content
5. Observe documented examples for "Default" and "Bordered" appear without and with borders respectively

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Improvements
 
 - Update styling for Tag component for consistency with U.S. Web Design System. ([#426](https://github.com/18F/identity-design-system/pull/426))
+- Hide collapsed Accordion component content while JavaScript loads. ([#434](https://github.com/18F/identity-design-system/pull/434))
 
 ### Dependencies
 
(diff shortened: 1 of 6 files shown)


zachmargolis (approved):

aduth:
Visual regression failure expected for revised Accordion documentation.

image

PR #435, opened 2024-03-20
Switch from HTML_CodeSniffer to Axe Puppeteer

aduth:
## 🛠 Summary of changes

Updates accessibility tests to use `@axe-core/puppeteer` instead of `HTML_CodeSniffer`, which has a much more thorough accessibility test suite and is more consistent with how we test accessibility in other Login.gov projects.

As part of this, updates were made to resolve preexisting accessibility issues:

- Update documented header markup to ensure [all content is contained in landmark regions](https://dequeuniversity.com/rules/axe/4.1/region), and for consistency with [documented USWDS component markup](https://designsystem.digital.gov/components/header/)
- Ensure that [scrollable code sections are keyboard-operable](https://dequeuniversity.com/rules/axe/4.8/scrollable-region-focusable), by overriding default Kramdown Markdown HTML conversion

## 📜 Testing Plan

1. `make build`
2. `node --test test/accessibility.test.js`


test/accessibility.test.js
```diff
@@ -42,21 +36,10 @@ describe('accessibility', () => {
...
-              resolve(window.HTMLCS.getMessages());
-            });
-          }),
-      );
-      for (const message of messages) {
-        if (message.type === MESSAGE_TYPE_ERROR) {
-          throw message;
-        }
-      }
+      const results = await new AxePuppeteer(page).withTags(['wcag2a', 'wcag2aa']).analyze();
```
aduth, line 39: This is "bare minimum". We could / should choose to omit `.withTags(['wcag2a', 'wcag2aa'])` which identifies several more "best practice" failures, but we can improve that iteratively.

```suggestion
      const results = await new AxePuppeteer(page).analyze();
```

docs/_plugins/scrollable_code_highlight.rb
```diff
@@ -0,0 +1,11 @@
+module Kramdown
```
zachmargolis, line 1: Do we need to explicitly load this anywhere so it can take effect? I'd expect a `_config.yml` change listing this file. Or does it get magically loaded somehow?

zachmargolis (approved):

docs/_plugins/scrollable_code_highlight.rb
aduth: Anything in `_plugins` is magically loaded by Jekyll, and this will override the module method from Kramdown referenced indirectly through Jekyll compilation.

I adapted this approach from similar plugins we use on some other sites like with how we [add design system classes to Markdown links in the brochure site](https://github.com/GSA-TTS/identity-site/blob/main/_plugins/content_typography.rb).

aduth:
Visual regression failure is catching updated markup for header component.

image

aduth:
>This is "bare minimum". We could / should choose to omit `.withTags(['wcag2a', 'wcag2aa'])` which identifies several more "best practice" failures, but we can improve that iteratively.

Some notes for posterity / future self after looking at this a bit: It's complicated by the fact that many code examples assume an isolated environment from sibling code examples so issues like [distinct `aria-label` for landmarks](https://dequeuniversity.com/rules/axe/4.3/landmark-unique) are flagged on pages where we show [multiple](https://github.com/18F/identity-design-system/blob/47fdeff32d476946ab4c4a0b031169a33449e068/docs/_components/side-navigation.md#L10) [variations](https://github.com/18F/identity-design-system/blob/47fdeff32d476946ab4c4a0b031169a33449e068/docs/_components/side-navigation.md#L29) of the same base markup. In practice we'd expect only one of those markups to be used on a site so that it wouldn't generate the issue.

Couple ideas for addressing this:

- Easier is to exempt certain "expected" violations for documentation pages after acknowledging it's not actually an issue
- Better would be to actually create that isolation with something like an `iframe`, similar to what was done in #418
   - Creating the frame page could be cumbersome to do manually.
   - As a solution, we could consider...
      - The dynamic injectable frame originally explored in #418 prior to ace14d0fb7ef97fd7c75b92f1cbb0be4b873ab34
      - A Jekyll plugin that [generates new content](https://jekyllrb.com/docs/plugins/generators/) in combination with a [custom tag](https://jekyllrb.com/docs/plugins/tags/) like `{% code_example %}`
         - It's not too hard to have a custom tag which renders Markdown and writes a new file, it's a little harder to place that rendered Markdown into an existing layout like the `base.html` layout

The "Easier" option might be the best bet for a short-term solution, after timebox exploration.

PR #438, opened 2024-04-22
Update Puppeteer, remove resize handling in visual regression specs

aduth:
## 🛠 Summary of changes

This pull request:

- Updates Puppeteer to the latest version
- Removes resize handling in visual regression specs

The intent is to try to resolve inconsistencies observed in visual regression failures in #437:

- As noted at https://github.com/18F/identity-design-system/pull/437#issuecomment-2069224563, the `fillImageToSize` helper doesn't work as intended, since it will only fill pixels which are missing outside both the expected width and height, but not either lacking height nor lacking width.
   - Ideally we don't want to tolerate any difference in dimensions. By git history, it's not entirely clear why this was added.
- In local testing, it seems that updating Puppeteer does have an impact in allowing the visual regression spec to pass, which could be due to differences in Chromium browser rendering. Unfortunately, due to how the visual regression test needs to change branches, `main` running an older version of Puppeteer can cause the failure to re-emerge. Performing the upgrade in a separate pull request allows the update to be made to `main` independent of the changes in #437.

## 📜 Testing Plan

Build should pass.

### DIFF
--- a/test/screenshot.test.js
+++ b/test/screenshot.test.js
@@ -18,37 +18,6 @@ const SNAPSHOT_DIRECTORY = 'tmp/screenshot/branches';
 const MAIN_SNAPSHOTS_DIRECTORY = join(SNAPSHOT_DIRECTORY, 'main');
 const BRANCH_SNAPSHOTS_DIRECTORY = join(SNAPSHOT_DIRECTORY, branch);
 
-/**
- * Resizes an image to the given dimensions.
- *
- * @param {import('pngjs').PNGWithMetadata} image
- * @param {number} width
- * @param {number} height
- *
- * @return {import('pngjs').PNG}
- */
-function fillImageToSize(image, width, height) {
-  if (image.width === width && image.height === height) {
-    return image;
-  }
-
-  const resizedImage = new PNG({ width, height, fill: true });
-  PNG.bitblt(image, resizedImage, 0, 0, image.width, image.height, 0, 0);
-
-  for (let y = image.height; y < height; y += 1) {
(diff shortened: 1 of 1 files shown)


aduth:
For posterity, I did create a "fixed" version of `fillImageToSize` if we need it again in the future:

```js
/**
 * Resizes an image to the given dimensions.
 *
 * @param {import('pngjs').PNGWithMetadata} image
 * @param {number} width
 * @param {number} height
 *
 * @return {import('pngjs').PNG}
 */
function fillImageToSize(image, width, height) {
  if (image.width === width && image.height === height) {
    return image;
  }

  const resizedImage = new PNG({ width, height, fill: true });
  PNG.bitblt(image, resizedImage, 0, 0, image.width, image.height, 0, 0);

  /**
   * @param {number} x
   * @param {number} y
   */
  function fillPixel(x, y) {
    // eslint-disable-next-line no-bitwise
    const index = (resizedImage.width * y + x) << 2;
    resizedImage.data[index] = 255; // Red
    resizedImage.data[index + 1] = 255; // Green
    resizedImage.data[index + 2] = 255; // Blue
    resizedImage.data[index + 3] = 255; // Alpha (Opacity)
  }

  for (let y = image.height; y < height; y++) {
    for (let x = 0; x < width; x++) {
      fillPixel(x, y + 1);
    }
  }

  for (let x = image.width; x < width; x++) {
    for (let y = 0; y < height; y++) {
      fillPixel(x, y);
    }
  }

  return resizedImage;
}
```

zachmargolis (approved):

PR #443, opened 2024-05-07
Restore, deprecate custom alert icons

aduth:
## 🛠 Summary of changes

Restores icons removed in #439, and marks them as deprecated in CHANGELOG.md pending the next major release.

These were removed previously because they were not internally referenced by the design system itself in its own styles, but I discovered that some downstream projects reference them directly ([example](https://github.com/18F/identity-dashboard/blob/09806682ede7edde503a7675385256b25c12330e/app/views/env/index.html.erb#L21-L32)), so removing them should be considered a breaking change.

## 📜 Testing Plan

Validate that `src/img/alerts` files removed in #439 are restored here.


nickttng (approved):
Good catch.

PR #442, opened 2024-05-07
Use USWDS icons with badge component

aduth:
## 🛠 Summary of changes

Updates Badge component to use USWDS icons.

This works toward removing our custom alert icons.

Related: #439, #440

## 📜 Testing Plan

1. Run locally as npm start
2. Go to http://localhost:4000/badges/
3. Observe badge icons

## 👀 Screenshots

Before|After
---|---
before|after

PR #445, opened 2024-05-07
Improve compatibility/precision for badge styles

aduth:
## 🛠 Summary of changes

Iterates on styles introduced in #442 targeting the badge icon.

The IdP has a [novel implementation](https://github.com/18F/identity-idp/pull/10065) of icon component which does not render markup as an `<svg />` element, but _does_ implement the equivalent as with the `usa-icon` class. Since it would be preferable (more precise) to implement the class selector vs. an element selector, this improves compatibility as well as improving the general implementation.

As part of the changes here, I dropped the selector which was deprecated as part of the work in #442, with the expectation this can be included in the next major release.

A patch release is not planned, and a temporary workaround will be added to the IdP instead.

## 📜 Testing Plan

Repeat testing plan from #442.

## 👀 Screenshots

There is not expected to be any visual impact of these changes, to be validated by the visual regression specs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Unreleased
 
+### Breaking Changes
+
+- Remove support for deprecation notices in v9.2.0:
+  - Badge component markup has been updated. ([#442](https://github.com/18F/identity-design-system/pull/442))
+
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


aduth:
I've been sitting on this in case I wanted to ship a minor release before the next major release, since it included a breaking change. Instead, I'm just going to keep the deprecation for now and address the specific fix, so it's not a breaking change.

I'll create a separate pull request closer to the next major release with removals for those deprecations.

PR #449, opened 2024-06-06
Update default alert styles to replace "Other" variant

aduth:
## 🛠 Summary of changes

Updates the Alert component styles to eliminate the custom `usa-alert--other` styles. This is intended to be backwards-compatible, where "Other" is intended to represent a stateless alert banner. This is already styled through USWDS as a slightly different shade of gray, and these changes help remove some confusion and extra code.

The concept of an "other" or default variant is already inherent to the idea of a BEM class "**B**lock", so it's redundant to have an explicit class. Technically it was already possible to use these styles with a `<div class="usa-alert">`, though it was never documented.

## 📜 Testing Plan

1. Go to http://localhost:4000/alerts/
2. Observe "Default" alert styles

## 👀 Screenshots

Before|After
---|---
image|image

### DIFF
--- a/docs/_components/alerts.md
+++ b/docs/_components/alerts.md
@@ -63,10 +63,10 @@ Visit the [USWDS Alerts component](https://designsystem.digital.gov/components/a
 {% endcapture %}
 {% include helpers/code-example.html code=example %}
 
-### Other
+### Default
 
 {% capture example %}
-<div class="usa-alert usa-alert--other">
+<div class="usa-alert">
   <div class="usa-alert__body">
     <p class="usa-alert__text">Banner text</p>
   </div>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Unreleased
 
+### Improvements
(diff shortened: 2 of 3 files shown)


nickttng (approved):
LGTM. Thanks for looking into this.

docs/_components/alerts.md
```diff
@@ -63,10 +63,10 @@ Visit the [USWDS Alerts component](https://designsystem.digital.gov/components/a
 {% endcapture %}
 {% include helpers/code-example.html code=example %}
 
-### Other
+### Default
```
nickttng, line 66: @aduth Are we calling the gray version as "Default" from now on, correct?

aduth: I think it makes sense to, since that is how these styles are applied.

Basically the USWDS CSS looks like:

```css
.usa-alert {
  // base
}

.usa-alert--info {
  // primary
}

// etc.
```

So this variant is the one without any BEM class modifiers (`--info`, etc), i.e. default.

